Issue #44, opened 2019-04-30
Do I need to download the tremendous weights again?


fanlyu:
I see the updates of the code, so do I need to download the tremendous weights again?

apsdehal:
Mostly not except for the new model. Features can stay the same but it will require some directory refactor.

fanlyu:
I try the old version weights that put vqa/train2014 and vqa/val2014 into train_val_2014.
Then I run the code with model pythia as
`python tools/run.py --tasks vqa --datasets vqa2 --model pythia --config configs/vqa/vqa2/pythia.yml`

But I got the error as below, how to slove it?

```
2019-04-30T17:16:15 INFO: Starting training...
2019-04-30T17:16:16 ERROR: 'Traceback (most recent call last):\n  File "/home/lvfan/anaconda3/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 138, in _worker_loop\n    samples = collate_fn([dataset[i] for i in batch_indices])\n  File "/home/lvfan/anaconda3/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 138, in <listcomp>\n    samples = collate_fn([dataset[i] for i in batch_indices])\n  File "/home/lvfan/pythia_new/pythia/tasks/multi_task.py", line 73, in __getitem__\n    item = self.chosen_task[idx]\n  File "/home/lvfan/pythia_new/pythia/tasks/base_task.py", line 154, in __getitem__\n    item = self.chosen_dataset[idx]\n  File "/home/lvfan/anaconda3/lib/python3.7/site-packages/torch/utils/data/dataset.py", line 81, in __getitem__\n    return self.datasets[dataset_idx][sample_idx]\n  File "/home/lvfan/pythia_new/pythia/tasks/base_dataset.py", line 49, in __getitem__\n    sample = self.get_item(idx)\n  File "/home/lvfan/pythia_new/pythia/tasks/vqa/vqa2/dataset.py", line 93, in get_item\n    return self.load_item(idx)\n  File "/home/lvfan/pythia_new/pythia/tasks/vqa/vqa2/dataset.py", line 127, in load_item\n    current_sample = self.add_answer_info(sample_info, current_sample)\n  File "/home/lvfan/pythia_new/pythia/tasks/vqa/vqa2/dataset.py", line 161, in add_answer_info\n    {"answers": answers, "tokens": sample_info["ocr_tokens"]}\nKeyError: \'ocr_tokens\'\n'
[...]

PR #46, opened 2019-05-06
Fix filepath and save state_dict for model for checkpoint finalize

vedanuj:
- Fixing bug in `pth_filepath` and `params_filepath` that was adding `save_dir` twice in the path.
- Fixing bug where the `self.trainer.model` when tried to save was throwing error. We need to save only the model `state_dict`. Fixes issue #44 error.

### DIFF
--- a/pythia/utils/checkpoint.py
+++ b/pythia/utils/checkpoint.py
@@ -36,14 +36,13 @@ class Checkpoint:
         self.config["log_foldername"] = self.ckpt_foldername
         self.ckpt_foldername = os.path.join(self.save_dir, self.ckpt_foldername)
         self.pth_filepath = os.path.join(
-            self.save_dir, self.ckpt_foldername,
+            self.ckpt_foldername,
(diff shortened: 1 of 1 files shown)


apsdehal:
I actually wanted to save the whole model itself for demo purposes. I guess it is fine with weights also. But, we probably should look into why this issue is exactly happening.

pythia/utils/checkpoint.py
```diff
@@ -36,15 +36,14 @@ def __init__(self, trainer):
         self.config["log_foldername"] = self.ckpt_foldername
         self.ckpt_foldername = os.path.join(self.save_dir, self.ckpt_foldername)
         self.pth_filepath = os.path.join(
-            self.save_dir, self.ckpt_foldername,
+            self.ckpt_foldername,
             self.ckpt_prefix + self.model_name +  "_final.pth"
         )
         self.params_filepath = os.path.join(
```
apsdehal: We should remove this line as it doesn't server any purpose now.

PR #47, opened 2019-05-06
Fix model loading from data parallel checkpoint

vedanuj:
Fix model loading when `.module` is not a prefix in attribute name. Currently it replaces non-prefix `.module` with empty strings. We should replace only the starting `module.` This bug occurred when a non data-parallel model was loaded.


apsdehal:
Is it still draft?

PR #48, opened 2019-05-07
Add tests for init_processor in BaseDataset class

vedanuj:
Adding tests for `init_processors` method in `pythia/tasks/base_dataset.py` testing both registry update and dataset class attributes update.


apsdehal:
Looks good. Thanks.

PR #55, opened 2019-05-20
Adding Image captioning to Pythia

vedanuj:
Adding image captioning task and Bottom Up Top Down attention model [link](https://arxiv.org/abs/1707.07998) into Pythia Framework. Model trained with the `butd.yml` config reaches close to the numbers reported in the paper. Also adds beam search decoding implementation. However this is not integrated to the standard flow yet. We can add something similar to `evalai` report generation for vqa datasets[needs discussion]. 

Currently we have added `bleu4` score from `nltk` library for `caption_accuracy` calculation.


pythia/modules/decoders.py
```diff
@@ -47,3 +49,78 @@ def forward(self, encoder_output, batch):
...
+    def __init__(
+        self,
+        attention_dim,
+        embed_dim,
+        decoder_dim,
+        vocab_size,
+        features_dim=2048,
+        dropout=0.5,
+    ):
+        super(CaptioningDecoder, self).__init__()
```
apsdehal: NIT: Be python 3 advocate and change this to `super().__init__()` maybe.

pythia/modules/decoders.py
```diff
@@ -47,3 +49,78 @@ def forward(self, encoder_output, batch):
...
+        self.language_model = nn.LSTMCell(
+            features_dim + decoder_dim, decoder_dim, bias=True
+        )
+        self.fc = nn.utils.weight_norm(
+            nn.Linear(decoder_dim, vocab_size)
+        )
+        self.init_weights()
+     
+    def init_weights(self):
+        self.fc.bias.data.fill_(0)
```
apsdehal: These should also be configurable from the configuration.

pythia/modules/decoders.py
```diff
@@ -47,3 +49,78 @@ def forward(self, encoder_output, batch):
...
+        self.fc.weight.data.uniform_(-0.1, 0.1)
+
+    def init_hidden_state(self, batch_size):
+        h = torch.zeros(batch_size, self.decoder_dim).to("cuda")
```
apsdehal: As mentioned above, please use `new` creation ops or use this and move to proper device.

pythia/modules/decoders.py
```diff
@@ -47,3 +49,78 @@ def forward(self, encoder_output, batch):
...
+        c = torch.zeros(batch_size, self.decoder_dim).to("cuda")
+        return h, c
+
+    def forward(self, image_features, caption_embeddings, state=None):
```
apsdehal: As discussed above, top-down attention class should be modified for this while being also usable for VQA and others. Decoder should technically be only generating the required text.

pythia/modules/losses.py
```diff
@@ -200,6 +201,32 @@ def forward(self, sample_list, model_output):
 
         return loss * targets.size(1)
 
+@registry.register_loss("cross_entropy")
+class CrossEntropyLoss(nn.Module):
```
apsdehal: Let's maybe call it something else. This is not a generic `CrossEntropyLoss` and is too specific to captioning and in general generation.

pythia/modules/metrics.py
```diff
@@ -49,6 +49,7 @@ def calculate(self, sample_list, model_output):
 import torch
 
 from pythia.common.registry import registry
+from nltk.translate.bleu_score import corpus_bleu
```
apsdehal: Let's not import it here. Import it only when your metric is initialized. This way this doesn't become a hard dependency.

pythia/modules/metrics.py
```diff
@@ -198,6 +199,86 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
         return value
 
 
+@registry.register_metric("caption_accuracy")
+class CaptionAccuracy(BaseMetric):
```
apsdehal: Is there some way it can be more generic and call handle any text strings and probabilities. I believe it requires minor changes. You will probably have to document in more detail.

Important: Please divide metrics into separate classes. This should be called `Bleu4` or `Blue4ForCaption` and you should implement other metrics separately so they are easily reusable.

pythia/modules/metrics.py
```diff
@@ -198,6 +199,86 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
...
+        references = list()
+        hypotheses = list()
+
+        # References
+        targets = sample_list.answers
+        for j, p in enumerate(targets):
+            img_caps = targets[j].tolist()
+            img_captions = list(
+                map(
+                    lambda c: [
```
apsdehal: Let's stay away from lambda functions. They are hard to test and can easily create issues. Can you move this to separate function and follow comments about readability as above?

pythia/modules/metrics.py
```diff
@@ -198,6 +199,86 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
...
+                    ],
+                    img_caps,
+                )
+            )
+            img_captions = [" ".join(c) for c in img_captions]
+            references.append(img_captions)
+
+        # Hypotheses
+        scores = model_output["scores"]
+        _, scores = torch.max(scores, dim=2)
```
apsdehal: NIT: Might be better to use `dim=-1` to be more generic.

pythia/modules/metrics.py
```diff
@@ -198,6 +199,86 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
...
+        scores = scores.tolist()
+        predictions = list()
+        for j, p in enumerate(scores):
+            for idx, v in enumerate(scores[j]):
```
apsdehal: NIT: Better variables names for readability.

pythia/modules/metrics.py
```diff
@@ -198,6 +199,86 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
...
 
 
+@registry.register_metric("caption_accuracy")
+class CaptionAccuracy(BaseMetric):
+    """Metric for calculating caption accuracy. Currently we are calculating BLEU4 Score.
+
+    **Key:** ``caption_accuracy``
+    """
+
+    def __init__(self):
```
apsdehal: Import nltk here.

pythia/tasks/captioning/coco/builder.py
```diff
@@ -0,0 +1,84 @@
...
+
+from pythia.common.registry import registry
+from pythia.tasks.base_dataset_builder import BaseDatasetBuilder
+from pythia.tasks.concat_dataset import PythiaConcatDataset
+
+from .dataset import COCODataset
+
+
+@registry.register_builder("coco")
+class COCOBuilder(BaseDatasetBuilder):
```
apsdehal: Isn't this very similar to VQA2Builder? Why not inherit that class?

pythia/tasks/captioning/coco/dataset.py
```diff
@@ -0,0 +1,152 @@
...
+
+from pythia.common.sample import Sample
+from pythia.tasks.base_dataset import BaseDataset
+from pythia.tasks.features_dataset import FeaturesDataset
+from pythia.tasks.image_database import ImageDatabase
+from pythia.utils.distributed_utils import is_main_process
+from pythia.utils.general import get_pythia_root
+
+
+class COCODataset(BaseDataset):
```
apsdehal: It probably will be better to inherit VQA2Dataset class here and use most of the similar code from it.

pythia/tasks/captioning/coco/dataset.py
```diff
@@ -0,0 +1,152 @@
...
+            return paths
+        else:
+            raise TypeError(
+                "Paths passed to dataset should either be " "string or list"
+            )
+
+    def __len__(self):
+        return len(self.imdb)
+
+    def try_fast_read(self):
```
apsdehal: Please use VQA2Dataset class so we can reuse most of the code.

pythia/tasks/captioning/coco/dataset.py
```diff
@@ -0,0 +1,152 @@
...
+    def format_for_evalai(self, report):
+        vocab = self.text_processor.vocab
+        captions = report.captions.tolist()
+        predictions = []
+        for idx, image_id in enumerate(report.image_id):
+            for idx_, el in enumerate(captions[idx]):
+                if el == vocab.EOS_INDEX:
+                    captions[idx] = captions[idx][:idx_]
+                    break
+            answer = [
```
apsdehal: Readability.

pythia/modules/metrics.py
vedanuj: I will move this logic to a caption processor.

pythia/modules/losses.py
vedanuj: Will rename to `CaptionCrossEntropyLoss`

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+
+        if len(complete_seqs_scores) == 0:
+            captions = torch.FloatTensor([0] * timesteps).unsqueeze(0)
+        else:
+            i = complete_seqs_scores.index(max(complete_seqs_scores))
+            captions = torch.FloatTensor(complete_seqs[i]).unsqueeze(0)
+
+        targets = sample_list.answers[:, 0]
+        sample_list.add_field("targets", targets)
+        sample_list.add_field("captions", captions)
```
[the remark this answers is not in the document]
vedanuj: `sample_list` requires a `targets` field otherwise it doesn't calculate loss. The `targets` is generated after we sort the captions in this batch in decreasing order. That is why I added it here. I can move the `captions` to `model_output`.

pythia/modules/attention.py
```diff
@@ -156,3 +156,25 @@ def forward(self, image_feat, question_embedding, image_locs=None):
                 masked_attention = self._mask_attentions(attention, image_locs)
 
         return masked_attention
+
+
+class CaptionAttention(nn.Module):
```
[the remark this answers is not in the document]
vedanuj: `ImageEmbedding` implementation uses `question_embedding` which we do not require for captioning. Also `AttentionLayer` uses modal combine layer for image and question embedding. We cannot use that as is here. I will look into if we can reuse any piece from there.

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+        data['image_features'] = data['image_features'][batch_size_t]
+        if 'state' in data:
+            h1 = data['state']['td_hidden'][0][batch_size_t]
+            c1 = data['state']['td_hidden'][1][batch_size_t]
+            h2 = data['state']['lm_hidden'][0][batch_size_t]
+            c2 = data['state']['lm_hidden'][1][batch_size_t]
+            data['state'] = {"td_hidden": (h1, c1), "lm_hidden": (h2, c2)}
+        return data
+
+    def beam_search(self, sample_list, k=5):
```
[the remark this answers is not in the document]
vedanuj: Will move this to utils. Currently `k` is not configurable from anywhere. We can add that as a config option.

pythia/modules/decoders.py
vedanuj: Do you mean keeping only the `Language LSTM` part as the decoder?

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+    def __init__(self, config):
+        super().__init__(config)
+        self.config = config
+        self._global_config = registry.get("config")
+        self._datasets = self._global_config.datasets.split(",")
+
+    def build(self):
+        self._build_word_embedding()
+        self._init_feature_encoders("image")
+        self._init_decoder()
```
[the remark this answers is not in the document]
vedanuj: I will try to modularize this and move the `TopDownAttention` part outside the decoder and make the decoder a configurable from configs. I will add these changes after I fix the other ones as this will require retraining the models and checking if we maintain the accuracy after the refactor.

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+
+        targets = (
+            sample_list.answers[:, 0, 1:]
+            if not teacher_forcing
+            else sample_list.text[:, 1:]
+        )
+        sample_list.add_field("targets", targets)
+        model_output = {"scores": scores}
+        del (data)
+        gc.collect()
```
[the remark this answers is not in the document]
vedanuj: I will remove this. Not necessary actually.

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+from pythia.common.registry import registry
+from pythia.modules.encoders import ImageEncoder
+from pythia.modules.decoders import CaptioningDecoder
+from torch import nn
+
+from .base_model import BaseModel
+
+
+@registry.register_model("butd")
+class BUTD(BaseModel):
```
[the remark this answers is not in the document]
vedanuj: Sounds good. Refactoring this will need retraining the current model we have. I will add these changes after I finish training with changes in this part.

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import gc
+import torch
+from pythia.common.registry import registry
+from pythia.modules.encoders import ImageEncoder
+from pythia.modules.decoders import CaptioningDecoder
+from torch import nn
+
+from .base_model import BaseModel
```
abhiskk: Is it possible to avoid relative import here: `from pythia.models.base_model import BaseModel`

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+    def _init_extras(self):
+        self.inter_model = None
+
+    def get_optimizer_parameters(self, config):
+        params = [
+            {"params": self.word_embedding.parameters()},
+            {"params": self.decoder.parameters()},
+            {
+                "params": self.image_feature_encoders.parameters(),
+                "lr": (config["optimizer_attributes"]["params"]["lr"] * 0.1),
```
abhiskk: I might be missing a detail here but why is the `lr` being multiplied by 0.1 here?

apsdehal: This layer comes from faster-rcnn and is being finetuned. That's why LR for fine-tuning is 0.1 times lower than the normal LR.

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+        return data
+
+    def beam_search(self, sample_list, k=5):
+        image_embedding_total  = self.process_feature_embedding("image", sample_list)
+
+        top_k_scores = torch.zeros(k, 1).to("cuda")
+        data = {}
+        data["texts"] = torch.LongTensor([[self.vocab.SOS_INDEX]] * k).to("cuda")
+        data["image_features"] = (
+            image_embedding_total.unsqueeze(1).expand(-1, k, -1, -1).squeeze(0)
```
abhiskk: A comment here would be helpful

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+            # Find next top k scores and words
+            if t == 0:
+                top_k_scores, top_k_words = scores[0].topk(k, 0, True, True)
+            else:
+                top_k_scores, top_k_words = scores.view(-1).topk(
+                    k, 0, True, True
+                )
+
+            # Convert to vocab indices
+            prev_word_inds = top_k_words / self.vocab_size
```
abhiskk: This will return a float, unless `top_k_words` is always divisible by `self.vocab_size`. Suggestion: `prev_word_inds = top_k_words // self.vocab_size`

pythia/models/butd.py
```diff
@@ -0,0 +1,294 @@
...
+                break
+            seqs = seqs[incomplete_inds]
+
+            data = self.get_data_beam_search_t(data, prev_word_inds[incomplete_inds])
+            top_k_scores = top_k_scores[incomplete_inds].unsqueeze(1)
+            data["texts"] = next_word_inds[incomplete_inds].unsqueeze(1)
+
+
+        if len(complete_seqs_scores) == 0:
+            captions = torch.FloatTensor([0] * timesteps).unsqueeze(0)
```
abhiskk: Use `torch.zeros` here.

pythia/modules/attention.py
```diff
@@ -156,3 +156,25 @@ def forward(self, image_feat, question_embedding, image_locs=None):
...
+    def __init__(self, features_dim, decoder_dim, attention_dim, dropout=0.5):
+        super(CaptionAttention, self).__init__()
```
abhiskk: nit: `super().__init__()`

pythia/modules/attention.py
```diff
@@ -156,3 +156,25 @@ def forward(self, image_feat, question_embedding, image_locs=None):
...
+        self.features_att = nn.utils.weight_norm(nn.Linear(features_dim, attention_dim))
+        self.decoder_att = nn.utils.weight_norm(nn.Linear(decoder_dim, attention_dim))
+        self.full_att = nn.utils.weight_norm(nn.Linear(attention_dim, 1))
+        self.relu = nn.ReLU()
+        self.dropout = nn.Dropout(p=dropout)
+        self.softmax = nn.Softmax(dim=1)
+
+    def forward(self, image_features, decoder_hidden):
+        att1 = self.features_att(image_features)
```
abhiskk: NIT: `att` -> `attn` is a better usage I think.

pythia/modules/attention.py
```diff
@@ -156,3 +156,25 @@ def forward(self, image_feat, question_embedding, image_locs=None):
...
+        att2 = self.decoder_att(decoder_hidden)
+        att = self.full_att(self.dropout(self.relu(att1 + att2.unsqueeze(1)))).squeeze(
+            2
+        )
+        alpha = self.softmax(att)
+        attention_weighted_encoding = (image_features * alpha.unsqueeze(2)).sum(dim=1)
+
+        return attention_weighted_encoding
```
abhiskk: NIT: be consistent in usage of `attn`, either stick to `attn` everywhere or `attention`

pythia/modules/decoders.py
```diff
@@ -2,6 +2,8 @@
 import torch
 from torch import nn
 
+from .attention import CaptionAttention
```
abhiskk: If possible use absolute import

pythia/models/butd.py
abhiskk: Maybe move this value also to `config`, 0.1 here looks like a magic number.

pythia/tasks/captioning/coco/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
abhiskk: The header message seems incomplete here

pythia/tasks/captioning/coco/dataset.py
```diff
@@ -0,0 +1,152 @@
...
+        if isinstance(paths, list):
+            return [self._get_absolute_path(path) for path in paths]
+        elif isinstance(paths, str):
+            if not os.path.isabs(paths):
+                pythia_root = get_pythia_root()
+                paths = os.path.join(pythia_root, self.config.data_root_dir, paths)
+            return paths
+        else:
+            raise TypeError(
+                "Paths passed to dataset should either be " "string or list"
```
abhiskk: Strong continuation looks odd here, I think this is a `black` artifact

pythia/tasks/captioning/coco/dataset.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def add_reference_caption(self, sample_info, sample):
+        reference_list = []
+        for reference in sample_info["reference_tokens"]:
+            text_processor_argument = {"tokens": reference}
+            processed_reference = self.text_processor(text_processor_argument)
+            reference_list.append(processed_reference["text"])
+
+        # Restrict to 5 reference captions
+        sample.answers = torch.stack(reference_list)[:5]
```
abhiskk: Make this a parameter instead of a constant

pythia/tasks/captioning/coco/__init__.py
vedanuj: Different files have different headers. Which is the correct one we should use?

pythia/modules/attention.py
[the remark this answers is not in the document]
apsdehal: But aren't you kind of conditioning on current word? Then question_embedding can be embedding of that word.

pythia/tasks/captioning/coco/__init__.py
apsdehal: This one works. It is short form of the original header used everywhere.

requirements.txt
```diff
@@ -10,3 +10,4 @@ PyYAML>=3.11
 pytest==3.3.2
 requests==2.21.0
 fastText
+nltk
```
apsdehal: Let's add the version number of nltk here.

vedanuj:
@apsdehal @abhiskk I have addressed most of the comments and restructured the code to reuse existing code as much as possible and I believe this is in a better state now. Would like to have your review on this. There are a few minor comments which I still have to address. Will do it soon.

vedanuj:
Updated the PR with all requested revisions.

configs/captioning/coco/butd.yml
```diff
@@ -0,0 +1,64 @@
...
+          out_dim: 1
+    image_feature_dim: 2048
+    embedding_dim: 300
+    image_feature_encodings:
+    - type: finetune_faster_rcnn_fpn_fc7
+      params:
+        bias_file: detectron/fc6/fc7_b.pkl
+        weights_file: detectron/fc6/fc7_w.pkl
+    inference:
+      type: greedy
```
apsdehal: Should we make it beam_search by default? If not, we should add some documentation for this

pythia/modules/layers.py
```diff
@@ -108,6 +111,37 @@ def forward(self, x):
         return logits
 
 
+class LanguageDecoder(nn.Module):
```
apsdehal: Maybe move this to `decoders.py` as it is a decode.

pythia/modules/layers.py
```diff
@@ -273,6 +309,45 @@ def forward(self, image_feat, question_embedding, context_embedding=None):
         return joint_feature
 
 
+class TopDownAttentionLSTM(nn.Module):
```
apsdehal: This should be moved to `attention.py`

pythia/modules/metrics.py
```diff
@@ -202,6 +202,59 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
         return value
 
 
+@registry.register_metric("caption_bleu4")
+class CaptionBleu4Metric(BaseMetric):
```
apsdehal: Are we providing support for other metrics? SPICE and ROGUE?

pythia/utils/beam_search.py
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
apsdehal: As discussed, let's move this to text_utils.

configs/captioning/coco/butd.yml
vedanuj: Currently the beam search works only with batch size 1. So adding it as the default inference type is not feasible during training for now. I will create a separate config file that can run beam search inference along with some comments.

pythia/modules/layers.py
vedanuj: I added it here since this is a `ClassifierLayer` type module and other classifier layers are also in this file.

pythia/modules/layers.py
vedanuj: I have added `TopDownAttentionLSTM` as a type of `ModalCombineLayer` so that we can reuse the `TopDownAttention` module when we process the feature and word embeddings. Does it still make sense to move to `attention.py` as this is only a type of layer?

pythia/modules/metrics.py
vedanuj: I have requested a separate PR #78 which adds a script for calculating metrics. If we are ok with importing the dependent libraries for calculating these metrics then we can add support for all of them in `metrics.py`.

vedanuj:
Thanks for the review. Should we add the tests in this PR or a separate PR?

apsdehal:
Separate is fine. Let's finish this one's comments and merge it. You should also make README, documentation and other changes in a separate PR.

pythia/modules/layers.py
apsdehal: Ok, keeping this here makes sense.

apsdehal:
This has landed into master 🎉. Pending TODOs:
- Update README and documentation updates #79 
- Add tests for captioning #80

PR #57, opened 2019-05-21
small bug in init of top down bottom up

shubhamagarwal92:
- small bug in init of top down bottom up

### DIFF
--- a/pythia/models/top_down_bottom_up.py
+++ b/pythia/models/top_down_bottom_up.py
@@ -15,7 +15,7 @@ from .base_model import BaseModel
 @registry.register_model("top_down_bottom_up")
 class TopDownBottomUp(BaseModel):
     def __init__(self, image_attention_model, text_embedding_models, classifier):
-        super(TopDownBottomUpModel, self).__init__()
+        super(TopDownBottomUp, self).__init__()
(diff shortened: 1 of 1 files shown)


pythia/models/top_down_bottom_up.py
```diff
@@ -15,7 +15,7 @@
 @registry.register_model("top_down_bottom_up")
 class TopDownBottomUp(BaseModel):
     def __init__(self, image_attention_model, text_embedding_models, classifier):
-        super(TopDownBottomUpModel, self).__init__()
+        super(TopDownBottomUp, self).__init__()
```
apsdehal: Can you actually change it to `super().__init__()`? Since, we are not supporting Pythia 2.x, we should do it in 3.x's way.

shubhamagarwal92: Done :)

apsdehal: Can you sign the CLA as well? Thanks.

shubhamagarwal92: I actually did! Not showing though

apsdehal: Thanks.

PR #60, opened 2019-05-23
CircleCI: Lint code for undefined names

cclauss:
Add [flake8](http://flake8.pycqa.org) tests to find Python syntax errors and undefined names.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 torch==1.0.1.post2
-torchvision>0.2
+torchvision==0.2.2
 tensorboardX==1.2
 numpy>=1.14
 tqdm>=4.19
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-torch==1.0.1.post2
+torch>=1.1.0
```
apsdehal: We are not yet ready to make this migration to 1.1.0. Can you revert this? It causes some issues in data parallel for us, so we need to figure it out properly before making this change.

cclauss: Reverted but that pegging that torch version causes the build to fail.

apsdehal:
Let's fix `torchvision==0.2.2`. This should fix the build.

apsdehal:
Let me merge this and fix the build issue on my end. Thanks.

bachagakshokh:
Yusaf Adfgh

PR #65, opened 2019-05-24
[feature] Abstract out configuration from trainer to enable custom training flows

meetshah1995:
Abstract out configuration from trainer to enable custom training flows

### DIFF
--- a/pythia/common/defaults/configs/base.yml
+++ b/pythia/common/defaults/configs/base.yml
@@ -1,5 +1,7 @@
 # Configuration for training
 training_parameters:
+    # Name of the trainer class used to define the training/evalution loop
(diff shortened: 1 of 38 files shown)


pythia/common/defaults/configs/base.yml
```diff
@@ -1,5 +1,7 @@
 # Configuration for training
 training_parameters:
+    # Name of the trainer class used to define the training/evalution loop
+    trainer_name: 'base_trainer'
```
apsdehal: Change this key to `trainer` instead of `trainer_name`? Rest looks good to me.

meetshah1995: Fixed in 294a0fa

PR #66, opened 2019-05-25
[misc] Fix some bugs, editorconfig, isort and black

apsdehal:
- See individual commits for more details
- Synchro and save checkpoint only in main process
- Fix a bug in early stopping
- Black and isort
- Add editorconfig for better cross editor compatibility
- Absolute imports instead of relative
- Logger now supports log all to log in all processes in distributed setting

### DIFF
--- a/pythia/common/trainer.py
+++ b/pythia/common/trainer.py
@@ -132,8 +132,7 @@ class Trainer:
                 rank, torch.cuda.get_device_name(self.local_rank)
             )
 
-            self.writer.write(device_info)
-            print(device_info)
+            self.writer.write(device_info, log_all=True)
 
         self.model = self.model.to(self.device)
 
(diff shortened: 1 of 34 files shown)


pythia/common/trainer.py
```diff
@@ -128,12 +128,13 @@ def load_model(self):
...
-            self.writer.write(
-                "CUDA Device {} is: {}".format(
-                    rank, torch.cuda.get_device_name(self.local_rank)
-                )
+            device_info = "CUDA Device {} is: {}".format(
+                rank, torch.cuda.get_device_name(self.local_rank)
             )
 
+            self.writer.write(device_info)
+            print(device_info)
```
abhiskk: NIT: If possible use logging here instead of print statements.

pythia/common/trainer.py
```diff
@@ -351,14 +351,16 @@ def _logistics(self, report):
...
 
         return should_break
 
     def _try_full_validation(self, force=False):
+        should_break = False
+
         if self.current_iteration % self.snapshot_interval == 0 or force:
-            self.writer.write("Evaluation time. Running on full "
-                              "validation set...")
+            self.writer.write("Evaluation time. Running on full " "validation set...")
```
dexter1691: nit: remove the two extra quotes here

apsdehal: Happens when I run black. Thanks.

pythia/common/trainer.py
apsdehal: Will be fixed, requires some changes in logger.

Issue #80, opened 2019-06-01
[tests] Add tests for Image Captioning


apsdehal:
For recent PR #55

apsdehal:
@vedanuj Any updates on this?

vedanuj:
I have to add the test for BeamSearch module. I will add it soon.

vedanuj:
#557 adds tests

PR #88, opened 2019-06-07
[tests] Adding tests for captioning losses and metrics

vedanuj:
Adding tests for caption losses and metrics. For issue #80 

Loss : `CaptionCrossEntropyLoss `
Metric : `CaptionBleu4Metric `


tests/modules/vocab.txt
```diff
@@ -0,0 +1,19 @@
...
+road
+riding
+motor
+bike
+the
+countryside
+back
+of
+motorcycle
+<unk>
```
apsdehal: Normally, <unk> is the first word in vocab.

vedanuj:
Added tests for `CaptionProcessor`. For `BeamSearch` it will need a bit of refactoring to be able to write the tests easily. I will do that together.

tests/modules/vocab.txt
vedanuj: Actually upon checking we need not add the `<unk>` token as `BaseVocab` class adds this token. Removed this.

apsdehal:
For beam search, are you going to do that in a separate PR?

PR #93, opened 2019-06-10
[feat] Add script for feature extraction from vmb

apsdehal:
- Allows one to extract features from vqa-maskrcnn-benchmark
- Also, saves info as separate npy files
- Downloads missing model and config files if not present

Importantly, we need to check the the extracting of extra info and if that is correct.

### DIFF
--- a/pythia/scripts/features/extract_features_vmb.py
+++ b/pythia/scripts/features/extract_features_vmb.py
@@ -1,4 +1,6 @@
 # Requires vqa-maskrcnn-benchmark to be built and installed
+# Category mapping for visual genome can be downloaded from
+# https://dl.fbaipublicfiles.com/pythia/data/visual_genome_categories.json
 import argparse
 import glob
 import os
@@ -24,6 +26,7 @@ class FeatureExtractor:
         "https://dl.fbaipublicfiles.com/pythia/detectron_model/detectron_model.yaml"
     )
     MAX_SIZE = 1333
+    MIN_SIZE = 800
     NUM_FEATURES = 100
 
     def __init__(self):
@@ -53,7 +56,14 @@ class FeatureExtractor:
             "--output_folder", type=str, default="./output", help="Output folder"
         )
         parser.add_argument("--image_dir", type=str, help="Image directory or file")
-
+        parser.add_argument(
+            "--feature_name", type=str, help="The name of the feature to extract",
+            default="fc6",
+        )
+        parser.add_argument(
+            "--confidence_threshold", type=float, default=0.2,
+            help="Threshold of detection confidence above which boxes will be selected"
+        )
         return parser
 
     def _build_detection_model(self):
@@ -78,8 +88,8 @@ class FeatureExtractor:
         im_size_min = np.min(im_shape[0:2])
         im_size_max = np.max(im_shape[0:2])
 
-        # Scale based on 800
-        im_scale = 800 / im_size_min
+        # Scale based on minimum size
+        im_scale = self.MIN_SIZE / im_size_min
 
         # Prevent the biggest axis from being more than max_size
         # If bigger, scale it down
@@ -93,13 +103,13 @@ class FeatureExtractor:
         return img, im_scale
 
     def _process_feature_extraction(
-        self, output, im_scales, feat_name="fc6", conf_thresh=0.2
+        self, output, im_scales, feature_name="fc6", conf_thresh=0.2
     ):
         batch_size = len(output[0]["proposals"])
         n_boxes_per_image = [len(boxes) for boxes in output[0]["proposals"]]
         score_list = output[0]["scores"].split(n_boxes_per_image)
         score_list = [torch.nn.functional.softmax(x, -1) for x in score_list]
-        feats = output[0][feat_name].split(n_boxes_per_image)
+        feats = output[0][feature_name].split(n_boxes_per_image)
         cur_device = score_list[0].device
 
         feat_list = []
@@ -150,7 +160,9 @@ class FeatureExtractor:
 
         with torch.no_grad():
             output = self.detection_model(current_img_list)
-        feat_list = self._process_feature_extraction(output, im_scales, "fc6", 0.2)
+        feat_list = self._process_feature_extraction(
+            output, im_scales, self.args.feature_name, self.args.confidence_threshold
+        )
         return feat_list
 
     def _chunks(self, array, chunk_size):
--- a/pythia/scripts/features/extract_features_vmb.py
+++ b/pythia/scripts/features/extract_features_vmb.py
@@ -1,4 +1,4 @@
-# Requires maskrcnn-benchmark to be built and installed
+# Requires vqa-maskrcnn-benchmark to be built and installed
 import argparse
 import glob
 import os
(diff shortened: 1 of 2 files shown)


pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
+# Requires maskrcnn-benchmark to be built and installed
```
vedanuj: Should this be `vqa-maskrcnn-benchmark`? We can be more explicit here so that it doesn't create any confusion. Maybe add a link to the gitlab repo url as well in comments.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+    CONFIG_URL = (
+        "https://dl.fbaipublicfiles.com/pythia/detectron_model/detectron_model.yaml"
+    )
+
+    def __init__(self):
+        self.args = self.get_parser().parse_args()
+        self.detection_model = self._build_detection_model()
+
+        if not os.path.exists(self.args.output_folder):
```
abhiskk: If you are using `exist_ok=True` you can remove the `os.path.exists` check.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+            print("Downloading model and configuration")
+            self.args.model_file = "./detectron_model.pth"
+            self.args.config_file = "./detectron_model.yaml"
+            download_file(self.MODEL_URL)
+            download_file(self.CONFIG_URL)
+
+    def get_parser(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            "--model_file", default=None, type=str, help="Detectron model file"
```
abhiskk: NIT: I have seen the patter `--model-file` to be more commonly used rather than `--model_file`. This is a tiny detail, feel free to ignore it.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+    def _image_transform(self, path):
+        img = Image.open(path)
+        im = np.array(img).astype(np.float32)
+        im = im[:, :, ::-1]
+        im -= np.array([102.9801, 115.9465, 122.7717])
+        im_shape = im.shape
+        im_size_min = np.min(im_shape[0:2])
+        im_size_max = np.max(im_shape[0:2])
+        im_scale = float(800) / float(im_size_min)
+        # Prevent the biggest axis from being more than max_size
```
abhiskk: NIT: empty line above this

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+        if np.round(im_scale * im_size_max) > 1333:
```
abhiskk: A comment here would be helpful

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+            im_scale = float(1333) / float(im_size_max)
```
abhiskk: I don't think a float casting is needed here.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+        im = cv2.resize(
```
abhiskk: Maybe use a torch transform here, or if this is consistent with transforms elsewhere ignore my comment.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+            im, None, None, fx=im_scale, fy=im_scale, interpolation=cv2.INTER_LINEAR
+        )
+        img = torch.from_numpy(im).permute(2, 0, 1)
+        return img, im_scale
+
+    def _process_feature_extraction(
+        self, output, im_scales, feat_name="fc6", conf_thresh=0.2
+    ):
+        batch_size = len(output[0]["proposals"])
+        n_boxes_per_image = [len(_) for _ in output[0]["proposals"]]
```
abhiskk: Rename `_` to a better name.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+            max_conf = torch.zeros((scores.shape[0])).to(cur_device)
+
+            for cls_ind in range(1, scores.shape[1]):
+                cls_scores = scores[:, cls_ind]
+                keep = nms(dets, cls_scores, 0.5)
+                max_conf[keep] = torch.where(
+                    cls_scores[keep] > max_conf[keep], cls_scores[keep], max_conf[keep]
+                )
+
+            keep_boxes = torch.argsort(max_conf, descending=True)[:100]
```
abhiskk: Rename 100 to a parameter.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def get_detectron_features(self, image_paths):
+        img_tensor, im_scales = [], []
+
+        for image_path in image_paths:
+            im, im_scale = self._image_transform(image_path)
+            img_tensor.append(im)
+            im_scales.append(im_scale)
+
+        current_img_list = to_image_list(img_tensor, size_divisible=32)
```
abhiskk: 32 looks like a magic number here

pythia/scripts/features/extract_features_vmb.py
apsdehal: This is mostly for our own internal use. `vqa-maskrcnn-benchmark` is not properly public yet.

pythia/scripts/features/extract_features_vmb.py
apsdehal: Would it matter? https://stackoverflow.com/questions/12834785/having-options-in-argparse-with-a-dash They are anyways converted to `_`. Do you know if there is a particular reason people use `-`.

apsdehal:
Addressed comments. It will useful to have review from @meetshah1995.

pythia/scripts/features/extract_features_vmb.py
meetshah1995: The underlying engine in torch transforms is PIL and the C2 transforms are cv2 based. To keep features exactly the same, it'd be better to use the same image processing engine.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,187 @@
...
+                cls_scores = scores[:, cls_ind]
+                keep = nms(dets, cls_scores, 0.5)
+                max_conf[keep] = torch.where(
+                    cls_scores[keep] > max_conf[keep], cls_scores[keep], max_conf[keep]
+                )
+
+            keep_boxes = torch.argsort(max_conf, descending=True)[:self.NUM_FEATURES]
+            feat_list.append(feats[i][keep_boxes])
+            bbox = output[0]["proposals"][i][keep_boxes].bbox / im_scales[i]
+            objects = torch.argmax(scores[keep_boxes], dim=1)
```
meetshah1995: Since the training is done on a subset of objects of VG, either releasing the subset annotation json files or having the class name in the detected objects as a key might be better.

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,187 @@
...
+            im_scales.append(im_scale)
+
+        # Image dimensions should be divisible by 32, to allow convolutions
+        # in detector to work
+        current_img_list = to_image_list(img_tensor, size_divisible=32)
+        current_img_list = current_img_list.to("cuda")
+
+        with torch.no_grad():
+            output = self.detection_model(current_img_list)
+        feat_list = self._process_feature_extraction(output, im_scales, "fc6", 0.2)
```
meetshah1995: Perhaps have `fc6` and `0.2` as parameters?

pythia/scripts/features/extract_features_vmb.py
```diff
@@ -0,0 +1,187 @@
...
+        img = Image.open(path)
+        im = np.array(img).astype(np.float32)
+        im = im[:, :, ::-1]
+        im -= np.array([102.9801, 115.9465, 122.7717])
+        im_shape = im.shape
+        im_size_min = np.min(im_shape[0:2])
+        im_size_max = np.max(im_shape[0:2])
+
+        # Scale based on 800
+        im_scale = 800 / im_size_min
```
meetshah1995: NIT: 800 could be a `MIN_SIZE` class attribute

meetshah1995:
Apart from some small comments, LGTM!

apsdehal:
Thanks everyone for reviews.

PR #102, opened 2019-06-19
[feature] Add CLEVR dataset

apsdehal:
Fixes #94 
- Adds CLEVR dataset and builder classes
- If not present CLEVR's data will be automatically downloaded and
extracted
- This will also create vocabularies
- Also adds CLEVR base config

PR #103, opened 2019-06-19
[feature] Add CNN LSTM model

apsdehal:
- A simple model that uses a CNN for image feature and LSTM for language. Then, it fuses them and passes them through a classifier
- Also, fixes accuracy metrics, adds more robust checks for some failure cases
- Adds a default config for cnn_lstm for CLEVR too.
- Key for the model is `cnn_lstm`
- Further, adds support for layers like `ConvNet`, `Flatten` and `UnFlatten` for easy use.


apsdehal:
Note: The tests should pass once #102 lands.

tests/models/test_cnn_lstm.py
```diff
@@ -0,0 +1,67 @@
...
+from pythia.utils.general import get_pythia_root
+
+
+class TestModelCNNLSTM(unittest.TestCase):
+    def setUp(self):
+        torch.manual_seed(1234)
+        registry.register("clevr_text_vocab_size", 80)
+        registry.register("clevr_num_final_outputs", 32)
+        config_path = os.path.join(
+            get_pythia_root(), "..", "configs", "vqa", "clevr", "cnn_lstm.yml"
```
mathfac: Maybe worth to implement method `get_pythia_related_path("../configs/vqa/clevr/cnn_lstm.yml")` to make life easier.

pythia/models/cnn_lstm.py
```diff
@@ -7,16 +7,44 @@
 from pythia.modules.layers import ConvNet, Flatten
 
 
+_TEMPLATES = {
```
mathfac: @asp, I thought about constants like `_QUESTION_VOCAB_SIZE_REG_NAME`, otherwise issue of typos still relevant.

PR #124, opened 2019-07-12
[fix] Make cudnn determinism true in case of fixed seed

### LINKED ISSUE #122: [bug] Fixing seed currently doesn't ensure true determinism in all cases
## 🐛 Bug

`torch.backends.cudnn.deterministic = True` also needs to be set.

apsdehal:
- By default seed is not set anything now
- If you specifically want to seed the training, you will have to
specify it now
- Custom DistributedSampler taken from maskrcnn-benchmark until upstream
PyTorch gets fixed
- Fixes #122

### DIFF
--- a/pythia/common/task_loader.py
+++ b/pythia/common/task_loader.py
@@ -3,11 +3,11 @@ import os
 
 import yaml
 from torch.utils.data import DataLoader
-from torch.utils.data.distributed import DistributedSampler
 
 from pythia.common.batch_collator import BatchCollator
 from pythia.common.test_reporter import TestReporter
 from pythia.tasks import MultiTask
+from pythia.tasks.samplers import DistributedSampler
 from pythia.utils.distributed_utils import get_world_size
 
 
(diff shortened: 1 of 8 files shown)


pythia/tasks/samplers.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
```
apsdehal: This has landed in master in https://github.com/pytorch/pytorch/pull/22479
But, since we are not yet on nightly build, I will keep this for now.

apsdehal:
Yup, that's one concern. According to the examples I have seen, maskrcnn-benchmark and pytorch's imagenet example has non-determinism as default behavior. I will give more thought and do more research on what exactly we should do.

pythia/common/task_loader.py
```diff
@@ -39,18 +39,20 @@ def get_test_reporter(self, dataset_type):
     def _load_task_config(self, task_name):
         directory = os.path.dirname(os.path.abspath(__file__))
         config_path = os.path.join(directory, "..", "tasks", task_name, "config.yml")
-        task_config = {}
+
+        self.task_config = {}
```
abhiskk: Maybe use a property for `task_config` as a wrapper over a private variable of `_task_config`

apsdehal:
The difference between seeded version and non-seeded version is insignificant atleast for pythia model. Merging this in.

PR #126, opened 2019-07-14
[fix] Extracting object features from gray image

HuaizhengZhang:
As the shape of gray image is 2D, 
https://github.com/facebookresearch/pythia/blob/0dd1d2f9dac3b2910aa2461e1915b27bc0bb0a4a/pythia/scripts/features/extract_features_vmb.py#L85

line 85 is as shown above will result in an IndexError
```
IndexError: too many indices for array
```


apsdehal:
This has landed into master, thanks a lot.

PR #127, opened 2019-07-15
[fix] Fix extract_features_vmb.py and extract_features.py to predict correct object classes

angelaslin:
This commit excludes the background class (index 0 of the scores vector) when computing the object class. This prevents the model from predicting the background class when the score of the background class is larger than the score computed using non-maximum suppression over non-background classes. This change also has the added benefit of aligning the object indices predicted by the model with the object mapping for Visual Genome (https://dl.fbaipublicfiles.com/pythia/data/visual_genome_categories.json).

### DIFF
--- a/pythia/scripts/features/extract_features_vmb.py
+++ b/pythia/scripts/features/extract_features_vmb.py
@@ -133,7 +133,9 @@ class FeatureExtractor:
             keep_boxes = torch.argsort(max_conf, descending=True)[:self.NUM_FEATURES]
             feat_list.append(feats[i][keep_boxes])
             bbox = output[0]["proposals"][i][keep_boxes].bbox / im_scales[i]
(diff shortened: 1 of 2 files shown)


angelaslin:
If you want to output the background class, do you also want to update the range in the for loop in extract_features_vmb.py (lines 126-131) to include the background class?

apsdehal:
Yup, thanks. So, let me write down the final proposal here:

- Instead of adding `no_background` flag, we add `background` flag.
- The change will also fix the bug of not starting from 1st index when calculating object label
- The `background` flag will start the loop on L121 from index 0 and also perform argmax including background class.
- A comment will added on the top regarding background class that object labels should be + 1 of what the object label is outputted and 0 is background class.

Thanks a lot for helping to make Pythia a better framework. :) Looking forward to your changes.

apsdehal:
Thanks a lot. This has landed into master.

PR #135, opened 2019-07-31
[feature] Add Nucleus Sampling decoding

ishan-maheshwari:
[feat] Add nucleus sampling implementation

[refactor] Arrange decoding strategies in different methods for readability

@dexter1691 @apsdehal


apsdehal:
A high level comment (I will review in detail later): You are missing tests and documentation for this PR. Can you add those?

ishan-maheshwari:
Yes. Working on it and I'll add them up soon.

vedanuj:
Thanks for the PR! Can you also add the metrics and if any improvements over using Beam Search with BUTD on COCO captioning task? You can use this script [here](https://github.com/facebookresearch/pythia/blob/master/pythia/scripts/coco/coco_caption_eval.py) to calculate all the metrics.

ishan-maheshwari:
Welcome! Yes, I'm going to work on creating tests, documentation and compiling quantitative and qualitative results this week.

pythia/models/butd.py
```diff
@@ -108,52 +109,92 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
+            embedding = self.word_embedding(pi_t)
+            attention_feature, _ = self.process_feature_embedding(
+                "image", sample_list, embedding[:, 0, :], batch_size_t=batch_size_t
             )
-            sample_list = beam_search.init_batch(sample_list)
+            output = self.classifier(attention_feature)
+            scores[:batch_size_t, t] = output
+
+        model_output = {"scores": scores}
+        return model_output
```
dexter1691: Maybe I am missing something, but shouldn't `forward_greedy` return the final caption too? `forward_beam_search` and `forward_nucleus_sampling` seems to be returning an additional key in the dictionary: `captions`.

pythia/models/butd.py
```diff
@@ -108,52 +109,92 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
+        # Stores the output probabilites.
+        scores = sample_list.answers.new_ones(
+            (
+                sample_list.answers.size(0),
+                self.text_processor.max_length,
+                self.vocab_size,
+            ),
+            dtype=torch.float,
+        )
+        if self.config["inference"]["type"] == "beam_search":
```
dexter1691: maybe replace these if conditions with a `getattr()` So something like:
`forward_fn = getattr(self, 'forward_' + self.config["inference"]["type"])`

pythia/utils/text_utils.py
```diff
@@ -289,3 +289,91 @@ def best_score(self):
...
+
+    def init_batch(self, sample_list):
+        setattr(
+            self,
+            "seq",
+            sample_list.answers.new_full(
+                (1, 1), self.vocab.SOS_INDEX, dtype=torch.long
+            ),
+        )
+        setattr(
```
dexter1691: what's the reason for initializing this here when you are replacing it in line 330
https://github.com/facebookresearch/pythia/blob/5370d1ba67958980dd3ee2d54564857720cfa37c/pythia/utils/text_utils.py#L330

pythia/utils/text_utils.py
```diff
@@ -289,3 +289,91 @@ def best_score(self):
...
+        top_m_words = top_m_words[:lastIndex]
+        # Zero value inside prev_word_inds because we are predicting a single stream of output.
+        prev_word_ind = torch.tensor([0])
+        # Get next word based on probabilities of top m words.
+        next_word_ind = top_m_words[torch.multinomial(self.top_m_scores, 1)]
+        # Add next word to sequence
+        self.seq = torch.cat(
+            [self.seq[prev_word_ind], next_word_ind.unsqueeze(1)], dim=1
+        )
+        # Check if sequence is complete
```
dexter1691: Redundant code in beam_search and here. Consider making this a function and calling it here.

pythia/utils/text_utils.py
```diff
@@ -289,3 +289,91 @@ def best_score(self):
...
+        if len(complete_inds) > 0:
+            self.complete_seq.extend(self.seq[complete_inds].tolist())
+            return True, data, 0
+
+        self.seq = self.seq[incomplete_inds]
+
+        # TODO: Make the data update generic for any type of model
+        # This is specific to BUTD model only.
+        data["texts"] = next_word_ind[incomplete_inds].unsqueeze(1)
+        h1 = data["state"]["td_hidden"][0][prev_word_ind[incomplete_inds]]
```
dexter1691: Same. Might make sense to avoid duplicate code?

pythia/models/butd.py
```diff
@@ -108,52 +109,92 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
+                "image", sample_list, embedding[:, 0, :], batch_size_t=batch_size_t
+            )
+            output = self.classifier(attention_feature)
+            # Compute Nucleus Sampling decoding
+            finish, data, batch_size_t = nucleus_sampling.sample(t, data, output)
+            if finish:
+                break
+
+        model_output = {"scores": scores}
+        model_output["captions"] = nucleus_sampling.get_caption()
```
dexter1691: `get_caption` is used for nucleus_sampling, while `best_score` is used for beam search. It'd be nice to make the function name consistent across sampling strategies. I like `get_caption`.

pythia/models/butd.py
ishan-maheshwari: The current version of butd.py only returns scores for greedy. I guess for greedy the caption generation is being handled somewhere else.

pythia/models/butd.py
ishan-maheshwari: Agreed. Will make the required changes.

pythia/models/butd.py
ishan-maheshwari: Makes sense. Will commit the required changes.

pythia/utils/text_utils.py
ishan-maheshwari: Right. Thinking of it now, top_m_scores isn't required to be defined as an attribute. Will fix.

pythia/utils/text_utils.py
ishan-maheshwari: Agreed. Should I make that function in utils/general.py?

pythia/models/butd.py
ishan-maheshwari: @dexter1691 you are right about this. I checked and greedy captioning shows error as it doesn't have 'captions' key in the dictionary. 

A short fix can be to reuse beam search module with beam_size = 1 to get captions. What do you think? Tagging @vedanuj (author).

ishan-maheshwari:
Adding Qualitative Comparisons between Nucleus Sampling and Beam Search.
[Qualitative Results.pdf](https://github.com/facebookresearch/pythia/files/3490151/Qualitative.Results.pdf)

apsdehal:
@ishan-maheshwari These analysis results look good. Can you also provide the captioning metrics as Vedanuj asked above, so we can compare better?

ishan-maheshwari:
Thanks! Working on getting metrics.

pythia/models/butd.py
apsdehal: So, here is a better way to do this:
Create a base class with `TextDecoder` which `BeamSearch` and `NucleusSampling` inherit. Greedy can just be beam search initialized with beam size = 1. You don't need separate forward for it.

Now, you don't need separate forward for all 3. Try to converge them in a pattern using encapsulation so that a single forward can be written for all 3. 

Seeing the code I see a lot of code getting repeated. Overall logic will remain same and all of the custom logic will go inside the decoders you implement. 

Let me know if this isn't clear to you. We can discuss more.

pythia/models/butd.py
```diff
@@ -108,52 +108,87 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
             )
-            sample_list = beam_search.init_batch(sample_list)
+            output = self.classifier(attention_feature)
+            scores[:batch_size_t, t] = output
 
+        model_output = {"scores": scores}
+        return model_output
+
+    def forward_beam_search(self, sample_list, scores):
+        beam_search = BeamSearch(
```
apsdehal: Now afterwards this will be initialized using constructor pattern, like how we construct the processors.

pythia/models/butd.py
```diff
@@ -108,52 +108,87 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
-        if self.config["inference"]["type"] == "beam_search":
-            model_output["captions"] = beam_search.best_score()
+        model_output["captions"] = beam_search.get_captions()
 
         return model_output
+
+    def forward_nucleus_sampling(self, sample_list, scores):
+        nucleus_sampling = NucleusSampling(
+                self.vocab, self.config["inference"]["params"]["sum_threshold"]
+            )
```
apsdehal: Black needs to be run on this code.

pythia/models/butd.py
```diff
@@ -108,52 +108,87 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
+        sample_list = nucleus_sampling.init_batch(sample_list)
+        batch_size = sample_list.image_feature_0.size(0)
```
apsdehal: `SampleList` has a function called `get_batch_size`. You don't need to do this.

pythia/models/butd.py
```diff
@@ -108,52 +108,87 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
-            model_output["captions"] = beam_search.best_score()
+        model_output["captions"] = beam_search.get_captions()
 
         return model_output
+
+    def forward_nucleus_sampling(self, sample_list, scores):
+        nucleus_sampling = NucleusSampling(
+                self.vocab, self.config["inference"]["params"]["sum_threshold"]
+            )
+        sample_list = nucleus_sampling.init_batch(sample_list)
```
apsdehal: Let's be consistent in naming when you create the base class. Call it `repackage_sample_list`

pythia/models/butd.py
apsdehal: With uniform structure this naming scheme will become consistent.

pythia/models/butd.py
```diff
@@ -108,52 +108,87 @@ def get_data_t(self, t, data, batch_size_t, prev_output):
...
+        # Stores the output probabilites.
+        scores = sample_list.answers.new_ones(
+            (
+                sample_list.answers.size(0),
+                self.text_processor.max_length,
+                self.vocab_size,
+            ),
+            dtype=torch.float,
+        )
+        return getattr(self, 'forward_' + self.config["inference"]["type"])(sample_list, scores)
```
apsdehal: No need of this after the inheritance structure. Decoder's encapsulation will take care of this.

pythia/utils/general.py
```diff
@@ -51,18 +50,13 @@ def clip_gradients(model, i_iter, writer, config):
 
 
 def ckpt_name_from_core_args(config):
-    seed = config["training_parameters"]["seed"]
-
-    ckpt_name = "{}_{}_{}".format(
+    return "%s_%s_%s_%d" % (
```
apsdehal: Please update and rebase based on latest master.

pythia/utils/general.py
```diff
@@ -204,3 +198,20 @@ def get_current_tensors():
...
+
+def find_complete_inds(decoder, next_word_inds):
+    incomplete_inds = []
+    for ind, next_word in enumerate(next_word_inds):
+        if next_word != decoder.vocab.EOS_INDEX:
+            incomplete_inds.append(ind)
+    complete_inds = list(set(range(len(next_word_inds))) - set(incomplete_inds))
+    return complete_inds, incomplete_inds
+
+def update_data(data, prev_word_inds, next_word_inds, incomplete_inds):
```
apsdehal: Move this to your own decoder. This isn't general enough to be here.

pythia/utils/general.py
```diff
@@ -204,3 +198,20 @@ def get_current_tensors():
                 print(type(obj), obj.size())
         except:
             pass
+
+def find_complete_inds(decoder, next_word_inds):
```
apsdehal: Same for this.

pythia/models/butd.py
apsdehal: Add this to modules decoders afterwards.

pythia/utils/text_utils.py
```diff
@@ -271,21 +261,93 @@ def search(self, t, data, scores):
...
+    def __init__(self, vocab, threshold=0.8):
+        self.vocab = vocab
+        self.vocab_size = vocab.get_size()
+        # Threshold for sum of probability
+        self.threshold = threshold
+        # Lists to store completed sequence
+        self.complete_seq = []
+
+    def init_batch(self, sample_list):
+        setattr(
```
apsdehal: Any specific reason to use `setattr`?

pythia/utils/text_utils.py
```diff
@@ -271,21 +261,93 @@ def search(self, t, data, scores):
...
+            .squeeze(0)
+        )
+        return sample_list
+
+    def sample(self, t, data, scores):
+        # Convert scores to probabilities
+        scores = torch.nn.functional.softmax(scores, dim=1)
+        # Sort scores in descending order and then select the top m elements having sum more than threshold.
+        # We get the top_m_scores and their indices top_m_words
+        if t == 0:
```
apsdehal: look into if it is possible to converge this into single line instead of if-else. Also try to reuse code from BeamSearch as much as possible after making a base decoder class.

pythia/utils/text_utils.py
```diff
@@ -271,21 +261,93 @@ def search(self, t, data, scores):
...
+            top_m_scores, top_m_words = scores[0].sort(
+                0, True
+            )
+        else:
+            top_m_scores, top_m_words = scores.view(-1).sort(
+                0, True
+            )
+
+        lastIndex = 0
```
apsdehal: Fix to snake_case here and other places.

pythia/utils/text_utils.py
```diff
@@ -271,21 +261,93 @@ def search(self, t, data, scores):
...
             i = self.complete_seqs_scores.index(max(self.complete_seqs_scores))
             captions = torch.FloatTensor(self.complete_seqs[i]).unsqueeze(0)
         return captions
+
+class NucleusSampling:
+    def __init__(self, vocab, threshold=0.8):
+        self.vocab = vocab
+        self.vocab_size = vocab.get_size()
+        # Threshold for sum of probability
+        self.threshold = threshold
```
apsdehal: Make this and wherever appropriate private variables.

pythia/utils/text_utils.py
```diff
@@ -271,21 +261,93 @@ def search(self, t, data, scores):
...
+        self.seq = self.seq[incomplete_inds]
+
+        # TODO: Make the data update generic for any type of model
+        # This is specific to BUTD model only.
+        data = update_data(data, prev_word_ind, next_word_ind, incomplete_inds)
+
+        return False, data, 1
+
+    def get_caption(self):
+        captions = torch.FloatTensor(self.complete_seq[0]).unsqueeze(0)
```
apsdehal: This needs to be on proper device instead of cpu until unless the last moment where it will be decoded.

ishan-maheshwari:
@apsdehal Noted all the points in review.

ishan-maheshwari:
Adding Quantitative Comparisons between Nucleus Sampling and Beam Search.
[Quantitative Results.pdf](https://github.com/facebookresearch/pythia/files/3516927/Quantitative.Results.pdf)

configs/captioning/coco/butd_nucleus_sampling.yml
```diff
@@ -0,0 +1,10 @@
+includes:
+- ../configs/captioning/coco/butd.yml
+model_attributes:
+  butd: &butd
+    inference:
+      type: nucleus_sampling
+      params:
+        sum_threshold: 0.8
+training_parameters:
+  batch_size: 1
```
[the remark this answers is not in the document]
ishan-maheshwari: I'm not sure about the training aspect. For now beam search in Pythia only worked for inference. I referred to beam search config for training parameters so that nothing breaks.

pythia/utils/text_utils.py
ishan-maheshwari: Yes. I guess most/all of the variables of self can be prefixed by underscore. Have done it.

pythia/utils/text_utils.py
ishan-maheshwari: I'm not sure what exactly is required here. Isn't decoding already complete when line 352 is executed?

pythia/utils/text_utils.py
ishan-maheshwari: This pattern was reused from Beam Search.

pythia/models/butd.py
ishan-maheshwari: I understand the base class approach. Should be pretty straightforward. Will update once done.

pythia/utils/general.py
ishan-maheshwari: Okay. Will make this change along with the base class change.

pythia/utils/text_utils.py
apsdehal: Hmm, but since `"seq"` is not a variable, directly assign it.

pythia/models/butd.py
ishan-maheshwari: I've made a base class that contains all the common code but I'm not able to figure out how to converge 3 forwards into one. An object of either BeamSearch or NucleusSampling needs to get created. A common forward can only do it through switch case or if-else. Even then, we can reduce a lot of redundancy. Should I make a common forward having conditional statements?

ishan-maheshwari:
@apsdehal to add documentation, I have created a directory for text_utils in docs folder and have added comments inline describing Nucleus Sampling as discussed with @dexter1691.

ishan-maheshwari:
Latest commit implements a common forward in butd.py, adopts an inheritance structure in text_utils.py and adds documentation for Nucleus Sampling.

pythia/utils/text_utils.py
ishan-maheshwari: This was a required conditional statement. Have reused as much as possible from base decoder class.

pythia/utils/text_utils.py
ishan-maheshwari: @apsdehal should this be considered resolved?

pythia/models/butd.py
ishan-maheshwari: This has been converted to a common forward with consistent naming convention.

pythia/models/butd.py
ishan-maheshwari: @apsdehal This has been changed to a common forward so can you have a look if any change is still required?

pythia/models/butd.py
ishan-maheshwari: Have reused this from beam search.

pythia/models/butd.py
ishan-maheshwari: @apsdehal @dexter1691 I've created a common forward but had to use a conditional statement because the calls to create a decoder object had different parameter requirements. If the differences were only limited to object class name then conditional statements could have been avoided.

pythia/models/butd.py
```diff
@@ -119,41 +119,36 @@ def forward(self, sample_list):
             dtype=torch.float,
         )
 
-        # For beam search inference. Currently beam seach for BUTD works only 
-        # with batch_size = 1 and should be used with run_type inference only.
-        # TODO : Implement batch beam search
-        if self.config["inference"]["type"] == "beam_search":
-            beam_search = BeamSearch(
+        if self.config["inference"]["type"] == "greedy":
+            decoder = BeamSearch(self.vocab, 1)
```
dexter1691: Can you pass self.vocab, self.config to the BeamSearch() and and NucleusSampling functions and extract the useful value inside the init function instead of the if condition here?

pythia/models/butd.py
```diff
@@ -119,41 +119,36 @@ def forward(self, sample_list):
...
         data, sample_list, timesteps = self.prepare_data(sample_list, batch_size)
         output = None
         batch_size_t = batch_size
         for t in range(timesteps):
             data, batch_size_t = self.get_data_t(t, data, batch_size_t, output)
-            if self.config["inference"]["type"] == "beam_search":
-                pi_t = data["texts"]
-            else:
-                pi_t = data["texts"][:, t].unsqueeze(-1)
+            pi_t = data["texts"]
```
dexter1691: Amm, the else condition is completely gone here. We don't require it now?

pythia/utils/text_utils.py
```diff
@@ -243,53 +296,113 @@ def search(self, t, data, scores):
...
+
+    Present text decoding strategies like beam search do not work well on open-ended
+    generation tasks (even on strong language models like GPT-2). They tend to repeat text
+    a lot and the main reason behind it is that they try to maximize likelihood, which is a
+    contrast from human-generated text which has a mix of high and low probability tokens.
+
+    Nucleus Sampling is a stochastic approach and resolves this issue. Moreover, it improves
+    upon other stochastic methods like top-k sampling by choosing the right amount of tokens
+    to sample from. The overall result is better text generation on the same language model.
+
```
dexter1691: Can you also add a link to the paper (with section number) so that people can refer to the paper with more details

pythia/models/butd.py
ishan-maheshwari: Nope. The else was required for greedy when it had its own code. But now since we're using Beam Search with beam_size = 1 for Greedy this else was no longer needed.

pythia/models/butd.py
ishan-maheshwari: Yeah about that. It would still require a conditional statement in Beam Search's init function because we want to use Greedy using Beam Search object only. Also, with that approach, how do we make a call for Greedy scenario without using if?

pythia/utils/general.py
```diff
@@ -52,18 +52,13 @@ def clip_gradients(model, i_iter, writer, config):
 
 
 def ckpt_name_from_core_args(config):
-    seed = config["training_parameters"]["seed"]
-
-    ckpt_name = "{}_{}_{}".format(
+    return "%s_%s_%s_%d" % (
```
apsdehal: Can you remove changes in this function? Seed is randomized in current master so it is not possible to resume easily in slurm when you add seed to the ckpt name.

pythia/models/butd.py
apsdehal: It would be better to change it at both places to be consistent with the API.

pythia/models/butd.py
dexter1691: Amm, can we not set the beam_length in the config to 1 for greedy decoding. This way, the design will be clean and we will be able to get rid of this if statement.

ishan-maheshwari: Yep. This should be doable, as long as we're okay with having mentions of 'beam' in greedy's config.

pythia/utils/general.py
ishan-maheshwari: Reverting the general.py to same as master. No changes in this file from my side anymore.

pythia/models/butd.py
ishan-maheshwari: I've tried to simplify it. Let me know if the latest version works?

pythia/models/butd.py
ishan-maheshwari: The get_batch_size method is causing dimension errors in scores array. So reverting back to original usage.

ishan-maheshwari:
Hi @apsdehal. CircleCI's tests have started to fail from 1st October because we're using an older version of PyTest that is causing errors now. Basically what's happening is - 

1. PyTest has a dependency called attrs that was updated on 1st October. That update is causing this problem.
2. PyTest released a new version 5.2.0 in which this error does not occur anymore.
3. Our CircleCI test is using an old version - 3.3.2 
4. So we need to change the PyTest version to 5.2.0

(source - https://stackoverflow.com/questions/58189683/typeerror-attrib-got-an-unexpected-keyword-argument-convert)

Can you have a look at this? The PR seems to be ready for merge but is blocked because of this.

apsdehal:
@ishan-maheshwari  Thanks for noticing this. You can make this change directly in your PR by modifying the requirements.txt file. CircleCI will pick it up.

ishan-maheshwari:
@dexter1691 @apsdehal The PR now has required testing for Nucleus Sampling. Would be great if you can have a look and see if it can be merged :)

apsdehal:
Thanks a lot for working on this amazing PR and the patience along with it. This has landed into master and would be available as part of v0.4 release.

PR #152, opened 2019-08-12
Add VQA EvalAI Accuracy metric

### LINKED ISSUE #109: [evaluation] Fix the gap between evaluation on EvalAI and in Pythia
## 🐛 Bug

Due to contractions and other stuff there is always a gap in between the validation performance on Pythia side and EvalAI side. This gap should probably be fixed at the end evaluation for the validation and test accuracies calculation.

## Expected behavior

Accuracies from Pythia and EvalAI should be same.

vedanuj:
Adding VQA Eval AI Accuracy `vqa_evalai_accuracy`  metric which is close to the accuracy calculation method in Eval AI. However this is slower compared to VQA Accuracy `vqa_accuracy` since it loops over and processes the answers before calculating the accuracy with the ground truth. Fixes #109.

### DIFF
--- a/pythia/modules/metrics.py
+++ b/pythia/modules/metrics.py
@@ -49,6 +49,7 @@ import collections
 import torch
 
 from pythia.common.registry import registry
+from pythia.utils.process_answers import ProcessAnswer
 
 
 class Metrics:
@@ -157,9 +158,7 @@ class BaseMetric:
 
         """
         # Override in your child class
-        raise NotImplementedError(
-            "'calculate' must be implemented in the child class"
-        )
+        raise NotImplementedError("'calculate' must be implemented in the child class")
 
     def __call__(self, *args, **kwargs):
         return self.calculate(*args, **kwargs)
@@ -194,10 +193,12 @@ class Accuracy(BaseMetric):
         output = model_output["scores"]
         expected = sample_list["targets"]
 
-        assert output.dim() <= 2, ("Output from model shouldn't have "
-                                   " more than dim 2 for accuracy")
-        assert expected.dim() <= 2, ("Expected target shouldn't have "
-                                     "more than dim 2 for accuracy")
+        assert (
+            output.dim() <= 2
+        ), "Output from model shouldn't have more than dim 2 for accuracy"
+        assert (
+            expected.dim() <= 2
+        ), "Expected target shouldn't have more than dim 2 for accuracy"
 
         if output.dim() == 2:
             output = torch.max(output, 1)[1]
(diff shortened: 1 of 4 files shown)


pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
-    "-",
-    ">",
-    "<",
-    "@",
-    "`",
-    ",",
-    "?",
-    "!",
-]
+class ProcessAnswer:
```
apsdehal: Maybe rename to `EvalAIAnswerProcessor`? Would it make sense for this to inherit `BaseProcessor`?

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
+    def __init__(self):
+        self.contractions = {
```
apsdehal: This should be a constant outside of the class `CONTRACTIONS`.

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
+            "y'all'dve": "y'all'd've",
+            "youd": "you'd",
+            "youd've": "you'd've",
+            "you'dve": "you'd've",
+            "youll": "you'll",
+            "youre": "you're",
+            "youve": "you've",
+        }
 
+        self.manual_map = {
```
apsdehal: Same here. Constant `NUMBER_MAP`

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
+            "six": "6",
+            "seven": "7",
+            "eight": "8",
+            "nine": "9",
+            "ten": "10",
+        }
+        self.articles = ["a", "an", "the"]
+        self.period_strip = re.compile("(?!<=\d)(\.)(?!\d)")
+        self.comma_strip = re.compile("(?<=\d)(\,)+(?=\d)")
+        self.punct = [
```
apsdehal: Same.

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
+        word = word.replace(",", "").replace("?", "").replace("'s", " 's")
+        return word.strip()
 
+    def process_punctuation(self, inText):
+        outText = inText
+        for p in self.punct:
+            if (p + " " in inText or " " + p in inText) or (
+                re.search(self.comma_strip, inText) is not None
+            ):
+                outText = outText.replace(p, "")
```
apsdehal: snake_case.

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
-        return 1
+    def word_tokenize(self, word):
+        word = word.lower()
+        word = word.replace(",", "").replace("?", "").replace("'s", " 's")
+        return word.strip()
 
+    def process_punctuation(self, inText):
+        outText = inText
+        for p in self.punct:
+            if (p + " " in inText or " " + p in inText) or (
```
apsdehal: snake_case

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
-        return 0.9
-    else:
-        return 1
+    def word_tokenize(self, word):
+        word = word.lower()
+        word = word.replace(",", "").replace("?", "").replace("'s", " 's")
+        return word.strip()
 
+    def process_punctuation(self, inText):
+        outText = inText
```
apsdehal: snake_case

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
-            re.search(comma_strip, inText) is not None
-        ):
-            outText = outText.replace(p, "")
-        else:
-            outText = outText.replace(p, " ")
-    outText = period_strip.sub("", outText, re.UNICODE)
-    return outText
+    def process_digit_article(self, inText):
+        outText = []
+        tempText = inText.lower().split()
```
apsdehal: Please check for snake_case everywhere.

apsdehal: I am aware that old code was also camelCase but it is right time to update this to be snake_case

pythia/utils/process_answers.py
```diff
@@ -5,244 +5,251 @@
...
+        return answer
 
-def process_digit_article(inText):
-    outText = []
-    tempText = inText.lower().split()
-    for word in tempText:
-        word = manual_map.setdefault(word, word)
-        if word not in articles:
-            outText.append(word)
+    def get_score(self, occurences):
```
apsdehal: min(occurrences * 0.3, 1)

pythia/utils/process_answers.py
```diff
@@ -289,7 +296,8 @@ def filter_answers(answers_dset, min_occurence):
         val_answers = json.load(open(val_annotation_file, "r"))["annotations"]
         answers = train_answers + val_answers
 
-    answer_list = filter_answers(answers, min_freq)
+    answer_processor = ProcessAnswer()
+    answer_list = answer_processor.filter_answers(answers, min_freq)
```
apsdehal: If we can add some simple tests for this, it would great.

pythia/modules/metrics.py
```diff
@@ -312,6 +313,79 @@ def calculate(self, sample_list, model_output, *args, **kwargs):
         return accuracy
 
 
+@registry.register_metric("vqa_evalai_accuracy")
+class VQAEvalAIAccuracy(BaseMetric):
```
apsdehal: I guess easiest way to use this would be to override command line args for metrics and then call specifically only val evaluation. We should ponder over better method to do this.

vedanuj:
`EvalAIAnswerProcessor` now inherits the `BaseProcessor` class. Addressed other comments about snake_case and tests.

pythia/modules/metrics.py
vedanuj: Do you mean adding the ability to specify a particular metric as a flag argument or just use the `vqa_evalai_accuracy` metric by default for validation on vqa?

PR #159, opened 2019-08-22
[fix] Fix detected objects in extract_features_vmb

### LINKED ISSUE #158: Minor bug in object detections in extract_features_vmb.py
## 🐛 Bug

I believe this was introduced by the #127, but there's a slight numpy indexing issue that causes the objects to still be detected incorrectly. 

The line in question is [here](https://github.com/facebookresearch/pythia/blob/12f67cd4f67499814bb0b3665ff14dd635800f63/pythia/scripts/features/extract_features_vmb.py#L165). It currently reads

```python
objects = torch.argmax(scores[keep_boxes][start_index:], dim=1)
```

However, `scores` is a tensor of `[num_objects,  object_classes]`, so `start_index` should be indexing the second dimension. The updated line should be

```python
objects = torch.argmax(scores[keep_boxes][:, start_index:], dim=1)
```

I can submit a pull request.

rmrao:
Fixes #158 

* [fix] Correctly ignore the background class

* [fix] Output the correct number of objects

### DIFF
--- a/pythia/scripts/features/extract_features_vmb.py
+++ b/pythia/scripts/features/extract_features_vmb.py
@@ -162,7 +162,7 @@ class FeatureExtractor:
             feat_list.append(feats[i][keep_boxes])
             bbox = output[0]["proposals"][i][keep_boxes].bbox / im_scales[i]
             # Predict the class label using the scores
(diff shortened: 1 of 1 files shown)


apsdehal:
Yup, you are right. Thanks for finding this.

PR #198, opened 2019-11-12
[fix] rename dataset_dts => dataset_gts in coco_caption_eval.py

ronghanghu:
The current `dataset_dts` in coco_caption_eval.py is actually ground-truth captions, so this name is confusing. Changing it to `dataset_gts` to avoid the confusion.


apsdehal:
Thanks. Landed into master.

PR #243, opened 2020-05-10
[feat,refactor] Resume from zoo, config, class refactor, more robust

apsdehal:
- Directly load from a zoo model with `resume_zoo` option
- Optionally, reset optimizers or counts
- Use `load_pretrained` to load using `zoo_override` otherwise correct
config must be known
- `resume_pretrained` added to clarify confusion on loading pretrained
mapping
- `pretrained_mapping` -> `pretrained_state_mapping`
- Complete some pending TODOs in checkpoint class

Test Plan:

Yet to be added
Also need to update some configs

### DIFF
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -1,3 +1,6 @@
+# Configuration version is useful in migrating older configs to new ones
+config_version: 1.0
+
 # Configuration for training
 training:
     # Name of the trainer class used to define the training/evalution loop
@@ -83,12 +86,19 @@ training:
 
     # Should a lr scheduler be used
     lr_scheduler: false
+
+    # DEPRECATED: Look at scheduler_attributes or
+    # Use PythiaScheduler directly instead
     # Steps for LR scheduler, will be an array of iteration count
     # when lr should be decreased
     lr_steps: []
+    # DEPRECATED: Look at scheduler_attributes or
+    # Use PythiaScheduler directly instead
     # Ratio for each lr step
     lr_ratio: 0.1
 
+    # NOTE: Have a look at newer scheduler available in MMF (such as AdamW) before
+    # using these options
     # Should use warmup for lr
     use_warmup: false
     # Warmup factor learning rate warmup
@@ -100,25 +110,6 @@ training:
     device: cuda
     # Local rank of the GPU device
     local_rank: null
-
-    # Use to load specific modules from checkpoint to your model,
-    # this is helpful in finetuning. for e.g. you can specify
-    # text_embeddings: text_embedding_pythia
-    # for loading `text_embedding` module of your model
-    # from `text_embedding_pythia`
-    pretrained_mapping: {}
-    # If using a pretrained model. Must be used with --resume_file parameter
-    # to specify pretrained model checkpoint. Will load only specific layers if
-    # pretrained mapping is specified in config
-    load_pretrained: false
-    # If resume is true, MMF will try to load automatically load
-    # last of same parameters from save_dir
-    resume: false
-    # `resume_file` can be used to load a specific checkpoint from a file
-    resume_file: null
-    # `resume_best` will load the best checkpoint according to monitored metric instead of
-    # the last saved ckpt
-    resume_best: false
 
     # If verbose dump is active, MMF will dump dataset, model specific
     # information which can be useful in debugging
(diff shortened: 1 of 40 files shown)


mmf/configs/defaults.yaml
```diff
@@ -220,3 +225,47 @@ distributed:
...
+checkpoint:
+    # If checkpoint.resume is true, MMF will try to load automatically load
+    # checkpoint and state from "current.ckpt" from env.save_dir
+    resume: false
+    # `checkpoint.resume_file` can be used to load a specific checkpoint from a file
+    # Can also be a zoo key
+    resume_file: null
+    # `checkpoint.resume_best` will load the best checkpoint according to
+    # training.early_stop.criteria instead of the last saved ckpt
+    resume_bert: false
```
vedanuj, line 239: `resume_best` typo

mmf/utils/checkpoint.py
```diff
@@ -106,8 +106,9 @@ def __init__(self, trainer):
             PathManager.mkdirs(self.models_foldername)
 
         self.save_config()
+
         self.repo_path = updir(os.path.abspath(__file__), n=3)
-        self.git_repo = None
+        self.git_repo = self.config.checkpoint.save_git_details
```
vedanuj, line 111: This won't work right? This will always set self.git_repo to `true` even when `git` module is not loaded (when the repo is not git). Check it's usage in L381

mmf/configs/defaults.yaml
```diff
@@ -220,3 +225,47 @@ distributed:
...
+    # `checkpoint.resume_pretrained` can be used in conjuction with `resume_file`
+    # or `resume_zoo` where you specify a checkpoint or .pth file to be loaded
+    # but it is mapped based on `checkpoint.pretrained_state_mapping`
+    resume_pretrained: false
```
vedanuj, line 243: I don't understand this. Why will this be used in "conjunction" with `resume_file` or `resume_zoo`? We don't need to load a pretrained model if we are loading from a resume_file.

mmf/configs/defaults.yaml
```diff
@@ -220,3 +225,47 @@ distributed:
...
+    # `checkpoint.resume_zoo` can be used to resume from a pretrained model provided
+    # in zoo. Value maps to key from zoo. `checkpoint.resume_file` has higher
+    # priority compared to `checkpoint.resume_zoo`.
+    resume_zoo: null
+    # `checkpoint.zoo_override` will override the current model config of trainer
+    # with what is provided from the zoo checkpoint and will load the model
+    # using .from_pretrained of the model passed
+    zoo_override: false
```
vedanuj, line 251: Better to rename it as `zoo_config_override` for clarity.

mmf/utils/checkpoint.py
```diff
@@ -154,137 +164,153 @@ def load_state_dict(self):
...
+                        key in formatted_attr
+                        and value in own_attr
+                        and formatted_attr.replace(key, "")
+                        == own_attr.replace(value, "")
+                    ):
+                        self.trainer.writer.write("Copying " + attr + " " + own_attr)
+                        own_state[own_attr].copy_(ckpt[attr])
+        self.trainer.writer.write("Pretrained model loaded")
+
+    def upgrade_state_dict(self, state_dict):
```
vedanuj, line 286: I think this `upgrade_state_dict` needs to be added in the `load_pretrained_model` method in L34 while loading.

mmf/configs/defaults.yaml
```diff
@@ -83,12 +83,17 @@ training:
 
     # Should a lr scheduler be used
     lr_scheduler: false
+
+    # DEPRECATED: Use MMFScheduler directly instead
```
vedanuj, line 87: Is MMFScheduler added in some earlier PR?

mmf/configs/defaults.yaml
vedanuj, line 239: `resume_best` typo

mmf/utils/checkpoint.py
apsdehal: I added this option so that you can override saving of git details even in the case of git repo.

mmf/configs/defaults.yaml
apsdehal: Both work. This is what `load_pretrained` used to be. This is a boolean value to be used in conjunction with `resume_file` or `resume_zoo`. Basically, in your change what you did via
`zoo_requirements` can be done via this as well.

mmf/configs/defaults.yaml
apsdehal: Updated.

mmf/utils/checkpoint.py
apsdehal: Fixed. Thanks.

mmf/configs/defaults.yaml
apsdehal: Updated.

mmf/configs/defaults.yaml
apsdehal: Fixed. Thanks.

mmf/utils/checkpoint.py
vedanuj: But won't this crash when `import git` fails?

apsdehal: You are right fixed this.

mmf/models/base_model.py
```diff
@@ -174,13 +173,11 @@ def __call__(self, sample_list, *args, **kwargs):
...
-            if isinstance(requirements, str):
-                requirements = [requirements]
-            for item in requirements:
-                download_pretrained_model(item, *args, **kwargs)
-        synchronize()
+        requirements = self.config.get("zoo_requirements", [])
+        if isinstance(requirements, str):
+            requirements = [requirements]
+        for item in requirements:
+            download_pretrained_model(item, *args, **kwargs)
```
vedanuj, line 180: Did you test this after removing the is_master() check here? Different processes will try to download the resource tar.

PR #245, opened 2020-05-11
[feature] HM prediction support, new predict cli cmd, support for csv predict file etc

vedanuj:
PR adds the following capabilities :
- New `mmf_predict` cli command for pypi. Can also be run as `python tools/predict.py .. `
- Support for saving predictions to a `csv` file in addition to default `json`
- New config keys `evaluation.predict` and `evalaution.predict_file_format`. Removes `evalai_inference`
- Predictions can be modified from models as well
- Add HM predict support 

Test Plan : 

```
mmf_predict config=projects/visual_bert/configs/hateful_memes/defaults.yaml dataset=hateful_memes model=visual_bert evaluation.predict=true evaluation.predict_file_format=csv
```

### DIFF
--- a/mmf/common/test_reporter.py
+++ b/mmf/common/test_reporter.py
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+import csv
 import json
 import os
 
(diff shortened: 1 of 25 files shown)


mmf/configs/defaults.yaml
```diff
@@ -135,6 +133,10 @@ training:
 evaluation:
     # Metrics for evaluation
     metrics: []
+    # Generate predictions in a file
+    predict: false
```
apsdehal, line 137: Shouldn't this by default be true in case of of predict command. It feels weird to ask user to pass this when they are explicitly saying predict.

mmf/trainers/base_trainer.py
```diff
@@ -494,8 +494,8 @@ def inference(self):
             self._inference_run("test")
```
apsdehal, line 494: here can you add `"test" in self.run_type`?

apsdehal: Actually, thinking more. I think run_type should be made a top level argument in defaults.yaml (at the same level as config). Why? Because it is being used by both evaluation and training and is not just about training.

mmf/common/test_reporter.py
```diff
@@ -152,6 +164,7 @@ def add_to_report(self, report):
         if not is_master():
             return
 
-        results = self.current_dataset.format_for_evalai(report)
+        results = self.current_dataset.format_for_prediction(report)
+        results = model.module.format_for_prediction(results, report)
```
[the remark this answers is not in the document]
vedanuj: Thanks for catching this.

mmf/configs/defaults.yaml
vedanuj: I will set this by default when people run `mmf_predict` cmd.

mmf/trainers/base_trainer.py
vedanuj: Ok I will move this to top level.

PR #251, opened 2020-05-13
[fix] Fix slurm scripts after config changes

vedanuj:
Summary: Fixing slurm scripts after configuration changes.

Differential Revision: D21558124


apsdehal:
Configuration upgrade didn't work?

PR #253, opened 2020-05-14
[docs] Add docs for configuration system, fix rtd build

apsdehal:
- Details hierarchy, overrides and other things
- Removes fasttext from rtd build to fix build and improve building time

Test Plan:
Built the documentation locally and tested the specific branch on RTD as well.


docs/source/notes/configuration.md
```diff
@@ -0,0 +1,194 @@
...
+
+```yaml
+includes:
+- path/to/first/yaml/to/be/included.yaml
+- second.yaml
+```
+
+The configs will override in the sequence of how they appear in the directive. Finally, the config parameters defined in the current config will override what is present in the includes. So, for e.g.
+
+First file, `a.yaml`:
```
vedanuj, line 117: In these examples can you use something that is related to our configs instead of mysql etc? Seems a bit odd.

docs/source/notes/configuration.md
```diff
@@ -0,0 +1,194 @@
+# Configuration System
+
+MMF relies on [OmegaConf](https://omegaconf.readthedocs.io/en/latest/) as its configuration system and adds some sugar on top of it. We have developed MMF as config-first framework. Most of the parameters/settings in MMF are configurable. MMF defines some default configuration settings for its system including datasets and models. User can then update this values via their own config or a command line dotlist.
```
vedanuj, line 3:
```suggestion
MMF relies on [OmegaConf](https://omegaconf.readthedocs.io/en/latest/) for its configuration system and adds some sugar on top of it. We have developed MMF as a config-first framework. Most of the parameters/settings in MMF are configurable. MMF defines some default configuration settings for its system including datasets and models. Users can then update these values via their own config or a command line dotlist.
```

PR #257, opened 2020-05-19
[tests] Fix device error in checkpoint test

vatsalg29:
There was an error in running pytest ./tests that it expected cuda object but received cpu object.


apsdehal:
i am not sure how you ran the test, but current doesn't test checkpoint in GPU (As you can check the CI tests are failing). You would need to move the model to the GPU if you want to test this on CUDA

vatsalg29:
I just ran pytest ./tests after installing from source using pip install --editable .

The **exact** error was - RuntimeError: Expected object of device type cuda but got device type cpu for argument #2 'other' in call to _th_equal
tests/utils/test_checkpoint.py:199: RuntimeError

After adding .cuda() the error was solved.

apsdehal:
@vatsalg29 Thank you for your contribution. This has landed in the master now. :)

PR #262, opened 2020-05-20
[fix,refactor] Add mmf_cli cmds, fix,speed up hm_convert, pip package

apsdehal:
- Python zipfile module uses Python interpretation to unzip files which
are encrypted; so improve performance we first use unzip command if
present otherwise fall back to zipfile implementation
- Bump the version number
- Add a new folder mmf_cli which will contain all cli commands in one
place but usually should symlink to tools folder
- Add proper __init__.py files for folder which were not getting
imported properly
- get_absolute_path now actually returns absolute path if the path is
not uri instead of directly returning the path. This fixes cases where
USER_DIR is set to "."
- If the folder is not git repository such as package folder git repo
creation in checkpoint used to throw error. Now, we catch that error and
just pass on it in case it is not a git repo
- Manifest file has been removed as package_data handles its uses
- Symlinks to packages that need to be included in mmf wheel have been
added in package subfolder so that they are easily includable as per
setuptools recommendation
- Automatically discovers the package data now
- Updates checksum for visual_bert coco
- Fix predict command for single GPU

Test Plan:
Tested first by publishing on test pypi


setup.py
```diff
@@ -88,7 +114,7 @@ def run(self):
 AUTHOR_EMAIL = "mmf@fb.com"
 DEPENDENCY_LINKS = []
 REQUIREMENTS = (fetch_requirements(),)
-EXCLUDES = ("data", "docs", "tests")
+EXCLUDES = ("data", "docs", "tests", "tests.*")
```
vedanuj, line 117: Not excluding directories?

setup.py
```diff
@@ -108,7 +134,9 @@ def run(self):
         name=DISTNAME,
         install_requires=REQUIREMENTS,
         include_package_data=True,
+        package_data={"mmf": fetch_package_data()},
         packages=setuptools.find_packages(exclude=EXCLUDES),
+        # data_files=fetch_data_files(),
```
vedanuj: Remove?

mmf_cli/__init__.py
```diff
@@ -0,0 +1 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
vedanuj, line 1: What is the advantage of adding these symlinks? Why not move the files entirely here?

mmf_cli/predict.py
```diff
@@ -0,0 +1 @@
+../tools/predict.py
```
vedanuj, line 1: Why can't we move run, predict and fb_run files here directly?

setup.py
```diff
@@ -69,6 +69,32 @@ def remove_specific_requirements(reqs):
...
+        if os.path.isdir(option):
+            files = []
+            for f in glob(os.path.join(option, "*")):
+                if os.path.isfile(f):
+                    files.append(f)
+                data_files += files
+    return data_files
+
+
+def fetch_package_data():
```
vedanuj, line 85: Just curious .. Why each files from folders need to be added?  Maybe some comments will be good for future maintainers.

mmf/projects
```diff
@@ -0,0 +1 @@
+../projects
```
vedanuj, line 1: Why these need to be added?

PR #265, opened 2020-05-20
[chores] Upgrade to Pytorch 1.5, torchvision 0.6

vedanuj:
- Upgrade to Pytorch 1.5, torchvision 0.6
- Update setup.py and hence rc version
- Update CircleCI for windows build test. Since windows instance we are using does not have GPU enabled we install cpu version of torch only.

Tests:
- Tests passing
- Tested with VisualBERT model training

### DIFF
--- a/mmf/version.py
+++ b/mmf/version.py
@@ -1,6 +1,6 @@
 import sys
 
-__version__ = "1.0.0rc5"
+__version__ = "1.0.0rc6"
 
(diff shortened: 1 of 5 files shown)


mmf/version.py
```diff
@@ -1,6 +1,6 @@
 import sys
 
-__version__ = "1.0.0rc3"
+__version__ = "1.0.0rc6"
```
apsdehal, line 3: Can you rebase?

PR #267, opened 2020-05-21
[feat,refactor] build,visualize dataset; refactor MultiDataset, fix HM

apsdehal:
- Adds builder utils for dataset and dataset loader
- setup_imports now have a check whether they were already setup and
always happen in the case of registry import
- Add typings file which contains multiple typings used build
- Adds visualize function to hateful memes dataset and to base dataset
class as well which can be overriden in the child datasets
- MultiDataset has been refactored to be more clean and use build utils
added in this PR.
- MultiDataset has been renamed to MultiDatasetLoader as it is actually
a loader and not a dataset
- Change accordingly in trainer
- Move setup imports function to env as that is the correct location
- Adds function load_yaml_with_defaults which will merge the yaml passed
with default MMF configuration file
- Adds option to image database's from_path function to not apply
transform
- Adds a new utils file visualize which contains visualize_images
function as of now
- HM dataset fix: Should not return target in case where label is not
present

Test Plan:

Test with regular visual_bert on hateful_memes command, pip install and
then test in jupyter notebook


mmf/utils/build.py
```diff
@@ -41,6 +44,116 @@ def build_model(config):
     return model
 
 
+def build_dataset(
```
vedanuj, line 47: Can these methods go to `utils/dataset.py` ? `build.py` file is getting too large and have build_* methods ranging from trainer class to classifier layers. Might be good to move the build methods to respective utils class.

mmf/utils/visualize.py
```diff
@@ -0,0 +1,24 @@
...
+        import matplotlib.pyplot as plt
+    except ImportError:
+        print(
+            "Visualization tools require matplotlib. "
+            + "Installing using pip install matplotlib."
+        )
+        raise
+    transform = torchvision.transforms.Compose(
+        [torchvision.transforms.Resize(size=size), torchvision.transforms.ToTensor()]
+    )
```
vedanuj, line 18: Do we always want to resize for visualization? Let's keep size `None` by default and then if someone wants to resize they can specify explicitly.

mmf/common/typings.py
```diff
@@ -0,0 +1,17 @@
...
+import torch
+
+from mmf.datasets.base_dataset import BaseDataset
+from mmf.datasets.base_dataset_builder import BaseDatasetBuilder
+
+DatasetType = Type[BaseDataset]
+DatasetBuilderType = Type[BaseDatasetBuilder]
+DictConfig = Type[omegaconf.DictConfig]
+DataLoaderAndSampler = Tuple[
+    Type[torch.utils.data.DataLoader], Union[torch.utils.data.Sampler, None]
```
vedanuj, line 15: Use `Optional[torch.utils.data.Sampler]` here.

mmf/common/typings.py
```diff
@@ -0,0 +1,17 @@
...
+]
+DataLoaderArgsType = Union[Dict[str, Any], None]
```
vedanuj, line 17: `Optional[Dict[str, Any]]`

mmf/datasets/builders/hateful_memes/dataset.py
```diff
@@ -77,12 +83,27 @@ def __getitem__(self, idx):
...
         return generate_prediction(report)
 
+    def visualize(self, num_samples=1, *args, **kwargs):
+        image_paths = []
+        random_samples = np.random.randint(0, len(self), size=num_samples)
+
+        for idx in random_samples:
+            image_paths.append(self.annotation_db[idx]["img"])
+
+        images = self.image_db.from_path(image_paths, use_transforms=False)
```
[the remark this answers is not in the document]
apsdehal: Right, I will fix this.

mmf/utils/build.py
apsdehal: Overall, I think keeping all of the build functions in a single file would be a better UX. Otherwise, user will need to know where exactly each function, even though may not be concerned about all utilities except build.py. I don't see an issue with file getting too large. (A lot of files in other projects are large as well, with IDE things should be easy to work with). I would like to keep all build functions together for now so that it is one stop-shop for anything you want to build.

mmf/utils/visualize.py
apsdehal: We need to keep this as torchvision takes tensor as a default which needs to have same dimension. This can anyways be override, I can add an assert that if size is None, the can only be one image.

PR #269, opened 2020-05-22
[fix,refactor] Upgrade to Python3.6+ syntax and pyupgrade hook

apsdehal:
- Upgrade to Python3.6+ syntax
- Use pyupgrade pre-commit hook to make sure everything stays on same
version

### DIFF
--- a/tools/sweeps/lib/slurm.py
+++ b/tools/sweeps/lib/slurm.py
@@ -34,7 +34,7 @@ def main(get_grid, postprocess_hyperparams, args):
         # launch training
         job_id = launch_train(args, config)
         if job_id is not None:
-            print("Launched {}".format(job_id))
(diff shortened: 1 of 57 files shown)


tools/sweeps/lib/slurm.py
```diff
@@ -51,7 +51,7 @@ def copy_all_python_files(source, snapshot_main_dir, code_snapshot_hash):
     """
     os.makedirs(snapshot_main_dir, exist_ok=True)
     destination = os.path.join(snapshot_main_dir, code_snapshot_hash)
-    assert not os.path.exists(destination), "Code snapshot: {0} alredy exists".format(
+    assert not os.path.exists(destination), "Code snapshot: {} alredy exists".format(
```
vedanuj, line 54: Is this correct?

PR #270, opened 2020-05-22
[chore] Add licenses to files missing it along with precommit hook

apsdehal:
- Adds pre-commit hook to automatically detect and add license if
missing
- Updates all of the current files


apsdehal:
I have checked internally and the single line works. I have added NOTICES file to attribute visual bert and block.bootstrap.pytorch repos. ViLBERT repo doesn't have a license.

PR #278, opened 2020-05-28
[feature] MMBT Interface for pretrained models

apsdehal:
- Add interface for MMBT which has classify function
- Create build_processors function for easy use with interfaces
- Allow no_tqdm in download
- Add tests for these interfaces

Test Plan:
Added tests for testing the pretrained model

### DIFF
--- a/mmf/models/interfaces/mmbt.py
+++ b/mmf/models/interfaces/mmbt.py
@@ -1,0 +1,86 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import os
+import tempfile
+from pathlib import Path
+from typing import Type, Union
+
+import torch
+import torchvision.datasets.folder as tv_helpers
+from PIL import Image
+from torch import nn
+
+from mmf.common import typings as mmf_typings
+from mmf.common.sample import Sample, SampleList
+from mmf.models.base_model import BaseModel
+from mmf.utils.build import build_processors
+from mmf.utils.download import download
+
+MMBT_GRID_HM_CONFIG_PATH = Path("projects/hateful_memes/configs/mmbt/defaults.yaml")
+ImageType = Union[Type[Image.Image], str]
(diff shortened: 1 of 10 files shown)


mmf/utils/build.py
```diff
@@ -225,3 +225,44 @@ def build_image_encoder(config, direct_features=False, **kwargs):
     else:
         module = ImageEncoder(config)
     return module.module
+
+
+def build_processors(
+    processors_config: mmf_typings.DictConfig, registry_key: str = None, *args, **kwargs
+) -> mmf_typings.ProcessorDict:
+    """Given a processor configs, builds the processors present and returns back
```
vedanuj, line 233:
```suggestion
    """Given a processor config, builds the processors present and returns back
```

mmf/models/interfaces/mmbt.py
```diff
@@ -0,0 +1,86 @@
...
+from mmf.utils.build import build_processors
+from mmf.utils.download import download
+
+MMBT_GRID_HM_CONFIG_PATH = Path("projects/hateful_memes/configs/mmbt/defaults.yaml")
+ImageType = Union[Type[Image.Image], str]
+PathType = Union[Type[Path], str]
+BaseModelType = Type[BaseModel]
+
+
+class MMBTGridHMInterface(nn.Module):
```
vedanuj, line 25: I would extend this class from `MMBT` and add only the extra methods. 

Ideally since every dataset can have different implementation for `classify` specific to a model, the better way to do it will be move dataset specific parts to the dataset classes and model specific parts to the model itself. That will help get rid of these extra interface class. We should discuss more about this design. I will approve this now since the PR needs to land urgently.

apsdehal: Sure. Let's discuss this separately outside of this PR.

PR #284, opened 2020-06-01
[feature, fix] Capture stdout in logger, fix logger initialization

vedanuj:
This PR addresses several logger related issues 

- Capture `stdout` by the logger. Adds a StreamToLogger class to write `stdout` to logger. Behavior can be overridden using config `training.stdout_capture`.
- Remove sending the writer/warning logs to stdout when we are capturing stdout logs in the logger
- Refactor out `build_config` as a separate method and decouple from `build_trainer`.
- Adds basic tests to check log files are created in proper place
- Fixes slurm script

### DIFF
--- a/mmf/utils/build.py
+++ b/mmf/utils/build.py
@@ -1,28 +1,53 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+
 import os
 import warnings
+from typing import Any, Type
 
 import torch
 from omegaconf import OmegaConf
 
 from mmf.common import typings as mmf_typings
 from mmf.common.registry import registry
+from mmf.utils.configuration import Configuration
 from mmf.utils.general import get_optimizer_parameters
 
 
-def build_trainer(configuration, *rest, **kwargs):
-    configuration.freeze()
+def build_config(
+    configuration: Type[Configuration], *args, **kwargs
+) -> mmf_typings.DictConfig:
+    """Builder function for config. Freezes the configuration and registers 
+    configuration object and config DictConfig object to registry.
 
+    Args:
+        configuration (Configuration): Configuration object that will be 
+            used to create the config.
+
+    Returns:
+        (DictConfig): A config which is of type Omegaconf.DictConfig
+    """
+    configuration.freeze()
     config = configuration.get_config()
     registry.register("config", config)
     registry.register("configuration", configuration)
 
+    return config
+
+
+def build_trainer(config: mmf_typings.DictConfig) -> Any:
+    """Builder function for creating a trainer class. Trainer class name
+    is picked from the config.
+
+    Args:
+        config (DictConfig): Configuration that will be used to create
+            the trainer.
+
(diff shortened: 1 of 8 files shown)


mmf/utils/logger.py
```diff
@@ -47,38 +49,49 @@ def __init__(self, config, name=None):
...
-        warnings_logger = logging.getLogger("py.warnings")
+        self.warnings_logger = logging.getLogger("py.warnings")
 
         # Set level
         level = config.training.logger_level
         self.logger.setLevel(getattr(logging, level.upper()))
         self._file_only_logger.setLevel(getattr(logging, level.upper()))
 
+        # Capture stdout to logger
+        self.stdout_logger = StreamToLogger(
```
apsdehal, line 60: There should be an option to disable stdout. I don't always want to override my sys.stdout.

apsdehal, line 60: Can we make all logger private variables? There is inconsistency here.

mmf_cli/run.py
```diff
@@ -30,9 +30,10 @@ def main(configuration, init_distributed=False, predict=False):
     registry.register("seed", config.training.seed)
     print(f"Using seed {config.training.seed}")
 
-    registry.register("writer", Logger(config, name="mmf.train"))
-
     trainer = build_trainer(configuration)
+
+    # Logger should be registered after config is registered in build_trainer
+    registry.register("writer", Logger(config, name="mmf.train"))
```
apsdehal, line 36: This is wrong. This will override the logger registered by the `build_trainer` function. We did it specifically so that if something needs to log before trainer is initialized, they are able to use it using this this custom logger with name `mmf.train`. This will be overridden by build_trainer function and will be replaced. So, this needs to go before `build_trainer`.

vedanuj: `build_trainer` doesn't register a logger. trainer.load() does. So this change doesn't affect that behavior. Before we initialize logger we need `config` to be registered, which is done in `build_trainer`. Otherwise when Logger initialization calls `get_env_key` it will get default config values. So currently logs are always saved in `./save/logs` irrespective of whether the `save_dir` is overidden or not. So either we can do it this way or move config registration out of `build_trainer` class. Third way is to use save_dir, log_dir directly from config passed to logger.

mmf/utils/logger.py
vedanuj: I will add an override function.

mmf/utils/build.py
```diff
@@ -1,28 +1,53 @@
...
+    """
+    configuration.freeze()
     config = configuration.get_config()
     registry.register("config", config)
     registry.register("configuration", configuration)
 
+    return config
+
+
+def build_trainer(config: mmf_typings.DictConfig) -> Any:
```
apsdehal, line 37: NIT: Return output should be Type[BaseTrainer]. Can be fixed later on once we finish trainer refactor.

PR #292, opened 2020-06-05
[feature, fix] added support for update_frequency

eundersander:
This PR adds support for config.training.update_frequency:
```
# Run update_frequency=K batches with batch_size=N, accumulate gradients, then do 
# one update (one optimizer step). The effect is a large effective batch size of 
# KxN (without incurring the memory overhead of setting batch_size to KxN).
update_frequency: 1
```
https://www.internalfb.com/intern/tasks/?t=66569918

With this change, we have to be more clear with naming. Amanpreet and I agreed on this:
- we have no particular term for a single forward/backward pass for a single batch
- "update" == "iteration" == one optimizer step (will be more than one batch when update_frequency > 1)

The update logic in `BaseTrainer._backward` has mostly been moved to `_check_start_update` and `_check_finish_update`.

There's also a fix for `MultiDatasetLoader.__len__`. The existing bug is only present for the case that the train set size is not a multiple of per-worker batch size, which is probably rare in practice.

Test Plan:
I did ad-hoc testing of BaseTrainer.train() with print statements and varying `[batch_size, update_frequency, train set size, max_updates, log_interval]`, including end-of-epoch and completion of the train run. 

I took at stab at comparing histograms of gradients used in each update. In theory, the we should see similar results between, say, `[batch_size=2, update_frequency=8]` and `[batch_size=16, update_frequency=1]` because these are the same effective batch size. In practice, the results were too noisy and thus inconclusive. I'm open to other suggestions for testing here.

### DIFF
--- a/mmf/trainers/base_trainer.py
+++ b/mmf/trainers/base_trainer.py
@@ -242,22 +242,25 @@ class BaseTrainer:
             if self.current_epoch > self.max_epochs:
                 break
 
-            for batch in self.train_loader:
+            epoch_num_batches = len(self.train_loader)
+
+            for epoch_batch_index, batch in enumerate(self.train_loader):
                 self.profile("Batch load time")
-                self.current_iteration += 1
-                self.writer.write(self.num_updates + 1, "debug")
+
+                self._check_start_update(epoch_batch_index, epoch_num_batches)
 
                 report = self._forward_pass(batch)
                 loss = self._extract_loss(report)
                 self._backward(loss)
-                should_break = self._logistics(report)
 
-                if self.num_updates > self.max_updates:
-                    should_break = True
+                should_break = self._check_finish_update(epoch_batch_index, report)
 
                 if should_break:
                     break
 
+                epoch_batch_index += 1
+
+            assert(epoch_batch_index == epoch_num_batches)
             # In distributed, each worker will complete one epoch when we reach this
             # as each worker is an individual instance
             self.current_epoch += get_world_size() - 1
@@ -278,16 +281,38 @@ class BaseTrainer:
         return report
 
     def _backward(self, loss):
-        self.optimizer.zero_grad()
+        # for update_frequency > 1, scale the loss so that accumulated gradients are the expected magnitude
+        loss = loss / self.config.training.update_frequency
         loss.backward()
-
-        if self.should_clip_gradients:
-            clip_gradients(self.model, self.num_updates, self.tb_writer, self.config)
-
-        self.optimizer.step()
-        self._run_scheduler()
-        self.num_updates += 1
         self.profile("Backward time")
+
+    def _check_start_update(self, epoch_batch_index, epoch_num_batches):
+        # When epoch_num_batches is not a multiple of the effective batch size, the last few batches
+        # will be discarded. I.e., they won't contribute to an update.
+        epoch_num_updates = epoch_num_batches // self.config.training.update_frequency
+        is_batch_discarded = (epoch_batch_index >= epoch_num_updates * self.config.training.update_frequency)
+        if is_batch_discarded:
+            return
+
+        is_first_batch_for_update = epoch_batch_index % self.config.training.update_frequency == 0
+        if is_first_batch_for_update:
+            self.current_iteration += 1
+            self.writer.write(self.num_updates + 1, "debug")
+            self.optimizer.zero_grad()
+
+    def _check_finish_update(self, epoch_batch_index, report):
+        should_break = False
+        is_last_batch_for_update = (epoch_batch_index + 1) % self.config.training.update_frequency == 0
+        if is_last_batch_for_update:
+            if self.should_clip_gradients:
+                clip_gradients(self.model, self.num_updates, self.tb_writer, self.config)
+            self.optimizer.step()
+            self._run_scheduler()
+            self.num_updates += 1
+            should_break = self._logistics(report)
+            if self.num_updates > self.max_updates:
+                should_break = True
+        return should_break
 
     def _extract_loss(self, report):
         loss_dict = report.losses
--- a/mmf/trainers/base_trainer.py
+++ b/mmf/trainers/base_trainer.py
@@ -260,7 +260,7 @@ class BaseTrainer:
 
                 epoch_batch_index += 1
 
-            assert(epoch_batch_index == epoch_num_batches)
+            assert epoch_batch_index == epoch_num_batches
             # In distributed, each worker will complete one epoch when we reach this
             # as each worker is an individual instance
             self.current_epoch += get_world_size() - 1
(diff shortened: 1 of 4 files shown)


mmf/trainers/base_trainer.py
```diff
@@ -278,16 +281,38 @@ def _forward_pass(self, batch):
...
 
     def _backward(self, loss):
-        self.optimizer.zero_grad()
+        # for update_frequency > 1, scale the loss so that accumulated gradients are the expected magnitude
+        loss = loss / self.config.training.update_frequency
         loss.backward()
+        self.profile("Backward time")
+
+    def _check_start_update(self, epoch_batch_index, epoch_num_batches):
+        # When epoch_num_batches is not a multiple of the effective batch size, the last few batches
```
eundersander, line 290: This is a weird case worth discussing. I decided to let the extra forward/backward passes run (see the train loop) and then we just ignore the results, which is wasteful computation. This behavior is identical to Pytorch Lightning's implementation of this feature (I think) [https://github.com/PyTorchLightning/pytorch-lightning/blob/82a20296e308a67c8d9202e4cbdf92a44b90b077/pytorch_lightning/trainer/training_loop.py].

mmf/utils/build.py
```diff
@@ -139,6 +139,7 @@ def build_dataloader_and_sampler(
             dataset_instance.dataset_name, dataset_instance.dataset_type
         ),
         num_workers=num_workers,
+        drop_last=False,  # see also MultiDatasetLoader.__len__
```
eundersander, line 142: `drop_last=False` is the default so I'm not changing behavior here. I'm just making our choice explicit.

mmf/trainers/base_trainer.py
```diff
@@ -278,16 +281,38 @@ def _forward_pass(self, batch):
...
+    def _check_finish_update(self, epoch_batch_index, report):
+        should_break = False
+        is_last_batch_for_update = (epoch_batch_index + 1) % self.config.training.update_frequency == 0
+        if is_last_batch_for_update:
+            if self.should_clip_gradients:
+                clip_gradients(self.model, self.num_updates, self.tb_writer, self.config)
+            self.optimizer.step()
+            self._run_scheduler()
+            self.num_updates += 1
+            should_break = self._logistics(report)
```
eundersander, line 312: Note this report (including the loss) is only from a single recent batch; it doesn't represent the entire update.

mmf/trainers/base_trainer.py
```diff
@@ -278,16 +280,49 @@ def _forward_pass(self, batch):
         return report
 
     def _backward(self, loss):
-        self.optimizer.zero_grad()
+        # for update_frequency > 1, scale the loss so that accumulated gradients are the
+        # expected magnitude
+        loss = loss / self.config.training.update_frequency
```
eundersander, line 285: See also https://github.com/PyTorchLightning/pytorch-lightning/blob/82a20296e308a67c8d9202e4cbdf92a44b90b077/pytorch_lightning/trainer/training_loop.py#L590

mmf/trainers/base_trainer.py
eundersander: Note also that this loss in this report is *not* the scaled-down loss (I mean the result of the division on L285).

mmf/trainers/base_trainer.py
```diff
@@ -242,22 +242,24 @@ def train(self):
...
-                should_break = self._logistics(report)
 
-                if self.num_updates > self.max_updates:
-                    should_break = True
+                should_break = self._check_finish_update(epoch_batch_index, report)
 
                 if should_break:
                     break
 
+                epoch_batch_index += 1
```
apsdehal, line 261: Does this need to be incremented here? Wouldn't enumerate automatically increment this?

mmf/trainers/base_trainer.py
```diff
@@ -278,16 +280,49 @@ def _forward_pass(self, batch):
...
+        # for update_frequency > 1, scale the loss so that accumulated gradients are the
+        # expected magnitude
+        loss = loss / self.config.training.update_frequency
         loss.backward()
+        self.profile("Backward time")
+
+    def _check_start_update(self, epoch_batch_index, epoch_num_batches):
+        # When epoch_num_batches is not a multiple of the effective batch size, the last
+        # few batches
+        # will be discarded. I.e., they won't contribute to an update.
```
apsdehal, line 292: Even after discarding, the forward and backward is still happening without `optimizer.step()` and `optimizer.zero_grad()`. Do we really need to skip them? what issues arise when you don't skip them?

mmf/trainers/base_trainer.py
apsdehal: Can we explicitly update the loss to the scaled down loss by assigning it directly? Wouldn't that fix the issue?

mmf/trainers/base_trainer.py
apsdehal: I understand it is hard to keep track of how many iterations among `update_frequency` iterations were completed. But maybe it is better to keep track of those explicitly rather than blindly dividing with `update_frequency`?

mmf/trainers/base_trainer.py
```diff
@@ -242,22 +242,24 @@ def train(self):
...
 
-            for batch in self.train_loader:
+            epoch_num_batches = len(self.train_loader)
+
+            for epoch_batch_index, batch in enumerate(self.train_loader):
                 self.profile("Batch load time")
-                self.current_iteration += 1
-                self.writer.write(self.num_updates + 1, "debug")
+
+                self._check_start_update(epoch_batch_index, epoch_num_batches)
```
apsdehal, line 250: Another possible way which we should discuss is to do following:
- Run a while loop
- Inside this while loop, run the for loop over `update_frequency` iterations.
- Use `iter(self.train_loader)` to get the next batch.
- If StopIteration error is found break the while loop
- Otherwise, follow the same flow with the batch you got now
- Do forward pass for each of the `update_frequency` items and keep accumulating reports and losses in a list
- When iterations finish run backward, stack the loss and divide by total number of elements and similarly for report have stack function.


This will keep most of the trainer logic same and give you ability of `update_frequency` easily.

mmf/trainers/base_trainer.py
eundersander: Can you elaborate? Are you saying you want to save the scaled-down loss in the report?

mmf/trainers/base_trainer.py
eundersander: Let's move this discussion to a chat channel and I'll post the resolution here.

mmf/trainers/base_trainer.py
eundersander: We're talking about restructuring the loop in a separate comment and offline chat. I'll revisit these comments after we've made a decision on the loop structure.

PR #307, opened 2020-06-08
[fix] Tensorboard logging folder fix and Database visualization fix

omkar-fb:
Summary:
 - Fixes log_folder attribute in TensorBoard logging
 - Fixes visualization when resize sizes are specified

### DIFF
--- a/mmf/trainers/base_trainer.py
+++ b/mmf/trainers/base_trainer.py
@@ -191,7 +191,7 @@ class BaseTrainer:
         self.tb_writer = None
 
         if self.training_config.tensorboard:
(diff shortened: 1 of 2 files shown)


vedanuj:
`size` param is taking `(224, 224)` by default. Shouldn't that be set to `None` as well? Otherwise this seems odd to check for `not None`

apsdehal:
@vedanuj It is not odd. Check the docstring of the function of what happens when size is None. By default, we should provide some value otherwise it is somewhat ambiguous for user to understand what they are expecting.

PR #315, opened 2020-06-10
[fix] Fix test reporter for vqa

vedanuj:
- Fixes test reporter when image_id has one dimension

Test Plan:

Test generating predictions for vqa

### DIFF
--- a/mmf/common/test_reporter.py
+++ b/mmf/common/test_reporter.py
@@ -159,9 +159,12 @@ class TestReporter(Dataset):
             if "question_id" in report:
                 report.question_id = gather_tensor(report.question_id).view(-1)
             if "image_id" in report:
-                _, enc_size = report.image_id.size()
                 report.image_id = gather_tensor(report.image_id)
-                report.image_id = report.image_id.view(-1, enc_size)
+                if len(report.image_id.shape) == 2:
+                    _, enc_size = report.image_id.size()
+                    report.image_id = report.image_id.view(-1, enc_size)
+                else:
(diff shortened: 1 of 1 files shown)


mmf/common/test_reporter.py
```diff
@@ -159,9 +159,12 @@ def add_to_report(self, report, model):
...
                 report.question_id = gather_tensor(report.question_id).view(-1)
             if "image_id" in report:
-                _, enc_size = report.image_id.size()
                 report.image_id = gather_tensor(report.image_id)
-                report.image_id = report.image_id.view(-1, enc_size)
+                if len(report.image_id.shape) == 2:
+                    _, enc_size = report.image_id.size()
+                    report.image_id = report.image_id.view(-1, enc_size)
+                else:
+                    report.image_id = report.image_id.view(-1)
```
apsdehal, line 167: Later, we should fix this properly by doing a gather all.

mmf/common/test_reporter.py
```diff
@@ -159,9 +159,12 @@ def add_to_report(self, report, model):
             if "question_id" in report:
                 report.question_id = gather_tensor(report.question_id).view(-1)
             if "image_id" in report:
-                _, enc_size = report.image_id.size()
                 report.image_id = gather_tensor(report.image_id)
-                report.image_id = report.image_id.view(-1, enc_size)
+                if len(report.image_id.shape) == 2:
```
ronghanghu, line 163: One should check the shape before (not after) `gather_tensor` (fixed in https://github.com/facebookresearch/mmf/pull/324)

PR #329, opened 2020-06-13
[feature] Add CPU tests to Actions

vedanuj:
- Add CPU tests to Github Actions. 
- Run a matrix : Ubuntu/Mac OS, Python 3.7, 3.8
- Proper caching of conda envs for both Linux and Mac
- Test results are uploaded as artifacts

- Linters(black, isort, flake8) will be added in a separate actions workflow for more clarity to developers where test fails. So not included with CPU tests here.
- CPU tests will be removed from Circle CI in subsequent PRs once we verify Actions is reliable 
- We cannot run tests for Windows yet as torch.distributed is not supported properly. Our tests can be fixed later if Windows tests are required and we can add `windows-latest` to the matrix.

Test Plan:
- Tested Actions on this branch

### DIFF
--- a/.github/workflows/cpu_test.yml
+++ b/.github/workflows/cpu_test.yml
@@ -1,0 +1,79 @@
+name: CPU Tests
+
+# This workflow is triggered on pushes to the repository.
+on: push
+
+jobs:
+  unittest:
+    strategy:
+      max-parallel: 8
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
+        python-version: [3.6, 3.7, 3.8]
+
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - name: Checkout master 🛎️
+      uses: actions/checkout@v2
+      # If you're using actions/checkout@v2 you must set persist-credentials to false in most
+      # cases for the deployment to work correctly.
+      with:
+        persist-credentials: false
+        ref: master
+        path: mmf_master
+
+    - name: Setup Conda Environment
+      uses: goanpeca/setup-miniconda@v1
+      with:
+        activate-environment: mmf
+        python-version: ${{ matrix.python-version }}
+        auto-update-conda: true
+        use-only-tar-bz2: true
+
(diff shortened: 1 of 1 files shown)


.github/workflows/cpu_test.yml
```diff
@@ -0,0 +1,83 @@
+name: CPU Tests
+
+on:
+  # Trigger the workflow on push to master or any pull request
+  push:
+    branches:
+      - master
```
apsdehal, line 7: What do you think about triggering it for all branches? Some of the branches aren't master but are important for the project. Also, this will help to understand CI failures before opening up the PR itself.

.github/workflows/cpu_test.yml
```diff
@@ -0,0 +1,83 @@
...
+        cd ${GITHUB_WORKSPACE}/mmf_master/tests
+        pytest --junitxml=artifacts/junit-${{ matrix.platform }}-python${{ matrix.python-version }}.xml -v .
+
+    - name: Upload Test Results
+      uses: actions/upload-artifact@v1
+      with:
+        name: pytest-results-${{ matrix.platform }}-python${{ matrix.python-version }}
+        path: mmf_master/tests/artifacts/junit-${{ matrix.platform }}-python${{ matrix.python-version }}.xml
+      # Use always() to always run this step to publish test results when there are test failures
+      if: ${{ always() }}
```
apsdehal, line 83: Is there a better way to do this?

.github/workflows/cpu_test.yml
vedanuj: Yeah we can do that.

.github/workflows/cpu_test.yml
vedanuj: This is the way I found from their [documentation](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#packaging-workflow-data-as-artifacts)

vedanuj:
We can fix the Windows tests in a separate PR.

PR #332, opened 2020-06-15
[feat, refactor] Trainer class refactor

vedanuj:
PR refactors the trainer class providing flexibility to extend from the BaseTrainer class and customize it. We also introduce a callback mechanism that can be executed at different events during training, validation or inference. Hence separates most of the logic of core training/optimization from other logistics.

- Create an abstract base trainer class, and core functionalities like loading dataset, model, optimizer, metrics are abstract methods that has to be implemented  by derived classes.  
- Create a new `MMFTrainer` class which keeps the same behavior as the current trainer but derived from the `BaseTrainer`.
- Move core logic of training loop, evaluation loop and prediction loop to Mixin classes and can be easily used in trainer classes.
- Create a callback mechanism similar to `PytorchLightning` with different types of events when callbacks can happen. With this it will be easier to have `PytorchLightning` supported trainer classes for `mmf`
- Callbacks are executed before or after events like training, validation, test, batch, epoch etc.
- Typing introduced in the trainer classes
- There can be some additional refactoring around how to include some additional callbacks like Early Stopping in the hooks, but that can be done  in future diffs.

Test Plan:
- Tested training, inference, validation runs, checked if events are happening in correct order with the new callback mechanism.

### DIFF
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -4,7 +4,7 @@ config_version: 1.0
 # Configuration for training
 training:
     # Name of the trainer class used to define the training/evalution loop
-    trainer: base_trainer
+    trainer: mmf_trainer
     # Seed to be used for training. -1 means random seed between 1 and 100000.
     # Either pass fixed through your config or command line arguments
     # Pass null to the seed if you don't want it seeded anyhow and
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -4,7 +4,7 @@ config_version: 1.0
 # Configuration for training
 training:
     # Name of the trainer class used to define the training/evalution loop
-    trainer: mmf_trainer
+    trainer: mmf
     # Seed to be used for training. -1 means random seed between 1 and 100000.
     # Either pass fixed through your config or command line arguments
     # Pass null to the seed if you don't want it seeded anyhow and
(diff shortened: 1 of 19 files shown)


mmf/configs/defaults.yaml
```diff
@@ -4,7 +4,7 @@ config_version: 1.0
 # Configuration for training
 training:
     # Name of the trainer class used to define the training/evalution loop
-    trainer: base_trainer
+    trainer: mmf_trainer
```
apsdehal, line 7: Can it be just called `mmf` instead of `mmf` trainer? Since, we are here, we should separately discuss how to provide options for trainers. Something like trainer attributes.

mmf/trainers/callbacks/checkpoint.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+from mmf.common.registry import registry
+from mmf.trainers.callbacks.base import Callback
+from mmf.utils.checkpoint import Checkpoint
+
+
+class CheckpointCallback(Callback):
```
apsdehal, line 8: Should we add tests for these?

mmf/utils/env.py
```diff
@@ -131,6 +131,9 @@ def setup_imports():
             importlib.import_module("mmf.models." + module_name)
         elif f.find("trainer") != -1:
             splits = f.split(os.sep)
+            folder_name = splits[-2]
```
apsdehal, line 134: This will need rebase.

apsdehal:
We should also write some docs on this separately after this PR.

Issue #328, opened 2020-06-13
Get bounding box information within Lorra model


Ra0k:
## ❓ Questions and Help

I am wondering if it is possible to get the bounding box information within the Lorra model (mmf/models/lorra.py) for both the image regions and the words. I would like to use the positions to develop a new model. If it is not possible, how should I proceed? Should I override the _process_feature_embedding_ method? Sorry, it is not crystal clear for me, how could I get the bounding box information exactly. 

Best regards,
David

apsdehal:
You can enable `return_features_info: true` to get bounding box features. Similarly check other options here: https://github.com/facebookresearch/mmf/blob/master/projects/m4c/configs/textvqa/defaults.yaml#L60

You can check which keys you have in the sample list using `sample_list.keys()` method.

Ra0k:
So far so good. I got access to "ocr_bbox_coordinates" once I got the dataloader fixed. I will make a pull request for that. On the other hand, the bounding box information for the image feats are still missing.

apsdehal:
Did you enable `return_features_info` to True, if you did and you should get it in the key "image_info_0".

Ra0k:
I used `sample_list.image_info_0.keys()` to print the keys and it gave me only the max features:
odict_keys(['max_features']). 

This is my config I added to the default one:
```
dataset_config:
  textvqa:
    use_ocr: true
    use_order_vectors: true
    return_features_info: true
    use_ocr_info: true
```

Ra0k:
@apsdehal, is it possible that the _use_order_vectors_ is overwritten somewhere on my end?

apsdehal:
Apologies for late reply @Ra0k. Let me have a look at my end and see what is the issue.

apsdehal:
I tested this and it seems like the lmdb doesn't have the information you need. I probably need to update the lmdb, looking into that now.

Ra0k:
@apsdehal thank you very much!

apsdehal:
@Ra0k Apologies for a late resolution on this one as it required extensive testing on our end as well as retraining of M4C model's best version. The PR #375 is now up and should be ready for your use.

Ra0k:
@apsdehal thank you very much for your help! 
However, in my end, I still can't access anything else but "max_features" within "image_info_0".
(I deleted the cache)

apsdehal:
Reinvestigating. I will reach back soon.

ronghanghu:
[...]

PR #335, opened 2020-06-16
[fix] Naming of OCR bounding box positions in utils/dataset level

Ra0k:
Related issue: #328

I ran into an error when I extended the default config file (mmf/projects/lorra/configs/textvqa/defaults.yaml) by the following properties:

```
dataset_config:
  textvqa:
    use_ocr: true
    use_order_vectors: true
    return_features_info: true
    use_ocr_info: true
```

The corresponding command:
```
python mmf/mmf_cli/run.py config=mmf/projects/lorra/configs/textvqa/defaults.yaml \
    datasets=textvqa \
    model=lorra \
    run_type=train_val
```

The solution aims to fix the naming issue.

### DIFF
--- a/mmf/utils/dataset.py
+++ b/mmf/utils/dataset.py
@@ -20,8 +20,8 @@ def build_bbox_tensors(infos, max_length):
 
     for idx, info in enumerate(infos):
         bbox = info["bounding_box"]
-        x = bbox["topLeftX"]
-        y = bbox["topLeftY"]
+        x = bbox.get("top_left_x", bbox["topLeftX"])
+        y = bbox.get("top_left_y", bbox["topLeftY"])
         width = bbox["width"]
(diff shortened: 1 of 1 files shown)


mmf/utils/dataset.py
```diff
@@ -20,8 +20,8 @@ def build_bbox_tensors(infos, max_length):
 
     for idx, info in enumerate(infos):
         bbox = info["bounding_box"]
-        x = bbox["top_left_x"]
```
apsdehal, line 23: I think this is due to older textvqa imdbs. Let's keep the backwards compatibility by changing this and the line below to:
```py
x = bbox.get("top_left_x", bbox["topLeftX"])
y = bbox.get("top_left_y", bbox["topLeftY"])
```

PR #336, opened 2020-06-17
[WIP] Adding VirTex model into MMF core.

kdexd:
Tracking all progress on adding complete support of VirTex models in MMF. VirTex comprises of a CNN "visual backbone" and a transformer decoder based "textual head". In terms of implementation, it would require adding image encoder and a text decode; followed by integrating with the registry and model zoo.

virtex model

**Canonical Implementation from Authors:** https://github.com/kdexd/virtex
**Paper:** https://arxiv.org/abs/2006.06666

Some of the modules I add here might be reusable for other supported models (e.g. `TorchvisionImageEncoder`), but require careful refactor in a backward compatible manner. Happy to get any feedback along the way!

### DIFF
--- a/mmf/modules/decoders.py
+++ b/mmf/modules/decoders.py
@@ -4,6 +4,7 @@ from torch import nn
 from torch.nn.utils.weight_norm import weight_norm
 
 from mmf.common.registry import registry
+from mmf.modules.embeddings import TextEmbedding
 
 
 class VisDialDiscriminator(nn.Module):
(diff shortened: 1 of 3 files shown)


mmf/modules/decoders.py
```diff
@@ -81,3 +82,124 @@ def forward(self, weighted_attn):
         state["lm_hidden"] = (h2, c2)
 
         return predictions
+
+
+class PyTorchTransformerDecoder(nn.Module):
```
apsdehal, line 87: Isn't this more of a `VirtextTransformerDecoder` given that Pytorch already has `nn.TransformerDecoder` (albeit without positional embedding)

PR #339, opened 2020-06-17
[feat] Adding LXMERT to list of available VL models for finetuning and pretraining

eltoto1219:
Adding model LXMERT.

To pretrain on VQA run:

`mmf_run config=projects/lxmert/configs/vqa2/pretrain.yaml run_type=train_val dataset=masked_vqa2 model=lxmert `

Added config files for GQA, COCO, VQA2.0, Visual Genome pretraining and VQA2 finetuning.


vedanuj:
I am adding GQA support in PR #349. Once that lands this can be tested with GQA pretraining and downstream. Note that this GQA dataset is using mmf features.

vedanuj:
Can you please rebase on master(GQA dataset has landed)?

eltoto1219:
I stopped pretraining because I could not get the ground truth answers to show up for either gqa or vqa2. I reproduced the original configs for the datasets,  but for some reason the "targets" attribute will not show up in the sampe_list. I was wondering if you guys could test this when you have time / if i should open an issue. the command I have been running is:

```
mmf_run config=projects/lxmert/configs/pretrain.yaml run_type=train_val dataset=masked_vqa2,masked_coco,masked_gqa model=lxmert
```

The other error I have been having is the visual_genome dataset is actually not running at all. Using the unchanged defaults.yaml for the dataset, I get an  error stating that the visual_genome dataset does not have the attribute "self.imdb". Since it inherits from the VQA2Dataset, I was wondering if this should instead be renamed to "self.annotation_db"

apsdehal:
@eltoto1219 We will take a look into GQA dataset. For VisualGenome, you are right. Try changing it to annotation_db and if it fixes it. Otherwise, we will fix both of them in single PR.

eltoto1219:
@apsdehal  Awesome thank you! I realized the problem with the answer targets not appearing in masked GQA + VQA is that after PR #349  which added support for GQA, I think the answer processors were moved outside of the dataset class, but there wasn't any code to add the final answer into the sample list. You can see my changes in the vqa2/ + gqa/ masked_dataset.py. 

The fix to the visual genome dataset worked, but I ran into another downstream problem: 
```
  File "/playpen/avmendoz/mmf/mmf/datasets/builders/vqa2/dataset.py", line 79, in load_item
    len(sample_info["question_tokens"]), dtype=torch.int
KeyError: 'question_tokens'
```
and then after a few iterations on the newly implemented GQA dataset, the folowing assertion was raised: 

```
  File "/playpen/avmendoz/mmf/mmf/datasets/processors/bert_processors.py", line 106, in _convert_to_indices

    assert len(input_ids) == self._max_seq_length
AssertionError
```

I am not quite sure how to address these moving forward

vedanuj:
Thanks for raising these issues. We will take a look at these.

projects/lxmert/configs/coco/masked.yaml
```diff
@@ -0,0 +1,58 @@
...
+            vocab_file: vocabs/vocabulary_100k.txt
+      masked_token_processor:
+        type: masked_token
+        params:
+          tokenizer_config:
+            type: bert-base-uncased
+            params:
+              do_lower_case: true
+          mask_probability: 0.15
+          max_seq_length: 20
```
vedanuj, line 53: Increase this to a larger value if some captions are longer than 20 tokens.

vedanuj:
Few comments and issues that I observed while testing this:

- Rebase  on latest master
- For earlier models we do not use answers during pretraining. That is why the answer_info was not added to GQA. Since LxMERT uses that it is fine to add that here.
- For Masked token processors increase the max_length to be more than maximum caption/question length for that individual dataset. In VisualBERT/ViLBERT we are using 128. This is the reason you are seeing the `AssertionError ` on GQA after few iterations.
- I didn't observe the error `KeyError: 'question_tokens'`. What is the combination of datasets that you used when you ran into that bug?
- I observed some dimension mismatch with `matched_loss` and `vision_loss`. You may want to recheck them once properly.
- Added some other comments to address

mmf/models/lxmert.py
```diff
@@ -0,0 +1,916 @@
...
+                    min(image_feature_variable.shape[1], bbox.shape[1]),
+                    dtype=torch.int).cuda()
+            bbox = bbox[:, :max_features.item(), ...]
+            image_h = getattr(image_info, "image_height", None)
+            image_w = getattr(image_info, "image_width", None)
+            if image_h is not None and image_w  is not None:
+                image_h = torch.tensor(image_h, dtype=torch.int)
+                image_w = torch.tensor(image_w, dtype=torch.int)
+                image_location = torch.zeros(tuple(bbox.shape))
+                image_location[:, :, 0] /= image_w[:, ...]
```
vedanuj, line 809: Change to `image_w[:, None]`. Similarly for others.

projects/lxmert/configs/coco/masked.yaml
dowhilefalse: Thanks for pointing this out. Original Implementation of LXMERT uses 20. So, we might want to keep it 20.

dowhilefalse:
Hi. thanks so much for the comments and suggestions! 
I am not sure what do you mean by mismatch with matched_loss and vision_loss. I checked them but didn't find issues.

vedanuj:
Thanks for rebasing to master. I will update some style changes and then this should be good to land.

vedanuj:
@eltoto1219 I have added a commit cleaning up some files and also fixing a few things. Please take a look. 

One thing we need to do before landing is remove duplication of code. You can use some of the methods from `transformers.modeling_bert` directly. Can you update the `lxmert.py` by removing methods that are already in `transformers.modeling_bert`?

dowhilefalse:
I updated most of the methods with transformers.modeling_bert and the rest are customized methods. I think it should be good to go after some testing.

mmf/configs/models/lxmert/defaults.yaml
```diff
@@ -0,0 +1,65 @@
...
+    r_layers: 5  # 0
+    visual_feat_dim: 2048
+    visual_pos_dim: 4
+    task_matched: true
+    task_mask_lm: true
+    task_obj_predict: true
+    task_qa: true
+    visual_losses:
+    - obj
+#     - attr
```
apsdehal, line 32: Remove?

projects/lxmert/configs/defaults.yaml
```diff
@@ -0,0 +1,81 @@
...
+    r_layers: 5  # 0
+    visual_feat_dim: 2048
+    visual_pos_dim: 4
+    task_matched: true
+    task_mask_lm: true
+    task_obj_predict: true
+    task_qa: true
+    visual_losses:
+    - obj
+#     - attr
```
apsdehal, line 31: Here as well?

projects/lxmert/configs/gqa/masked.yaml
```diff
@@ -0,0 +1,57 @@
+dataset_config:
```
apsdehal, line 1: Can we avoid replication of configs among each other? Same thing is replicated multiple times which makes it hard to process what exactly are the exact changes from one config to other. Rely on default configs and includes as much as possible to avoid replication.

mmf/models/lxmert.py
```diff
@@ -0,0 +1,800 @@
...
+    def __init__(self, config):
+        super().__init__(config)
+
+    @classmethod
+    def config_path(cls):
+        return "configs/models/lxmert/pretrain.yaml"
+
+    # Backward compatibility
+    @classmethod
+    def format_state_key(cls, key):
```
apsdehal, line 646: Is this needed? This was the case for older models, LXMERT doesn't need it.

mmf/models/lxmert.py
```diff
@@ -0,0 +1,800 @@
...
+        image_info = getattr(sample_list, "image_info_0", {})
+        image_dim_variable = getattr(image_info, "max_features", None)
+        image_feature_variable = getattr(sample_list, "image_feature_0", None)
+        if image_feature_variable is None:
+            raise Exception("object features must be provided for lxmert")
+        bbox = getattr(image_info, "bbox", None)
+        if bbox is None:
+            raise Exception("bounding boxes must be provided for lxmert")
+        else:
+            bbox = torch.tensor(bbox)
```
apsdehal, line 679: I believe this was refactored into a processor?

mmf/models/lxmert.py
```diff
@@ -0,0 +1,800 @@
...
+                image_h = torch.tensor(image_h, dtype=torch.int)
+                image_w = torch.tensor(image_w, dtype=torch.int)
+                image_location = torch.zeros(tuple(bbox.shape))
+                image_location[:, :, 0] /= image_w[:, None]
+                image_location[:, :, 1] /= image_h[:, None]
+                image_location[:, :, 2] /= image_w[:, None]
+                image_location[:, :, 3] /= image_h[:, None]
+            else:
+                image_location = bbox
+            image_location_variable = image_location.cuda()
```
apsdehal, line 696: Use `.to(device)` with proper device from registry. This will not allow LXMERT to run on CPU. Please take care of this at other cuda calls as well.

apsdehal: Alternatively, you can `torch.new_*` to create tensors with same dtype as input to avoid figuring out device manually.

mmf/models/lxmert.py
```diff
@@ -0,0 +1,800 @@
...
+
+        # aux data
+        image_label_variable = getattr(sample_list, "image_labels", None)
+        if image_label_variable is not None:
+            image_label_variable = torch.tensor(image_label_variable, dtype=torch.long)
+            image_label_variable = image_label_variable[:, : max_features.item(), ...]
+            image_label_variable = image_label_variable.unsqueeze(-1).cuda()
```
apsdehal, line 703: Can't this possible be done in the dataset processor itself?

dowhilefalse:
Hi! Thanks for the feedback.
I changed the config files and now they don't have overlaps. 
I changed .cuda to proper device call in lxmert.py. I tested locally and it can run on CPU.


another issue is that visual_genome run report key 'question_tokens' not found error.

I checked datatsets/builders/vqa2/dataset.py and found lines 80 might have some issue.
current_sample.text_len =torch.tensor(
            len(sample_info["question_tokens"]), dtype=torch.int
         )
this might have some issue since question_tokens might not be present.

can you please check it. Thanks so much!

vedanuj:
@dowhilefalse Thanks for updating with the changes. Can you please rebase your branch on master? Lot of changes are not related to this PR.

vedanuj:
@eltoto1219  I pushed the latest master changes to your master branch and then it denied permission to me for any further pushes. Can you check if there is any permission issue blocking pushes to your repo? If it can't be fixed I will open a new PR with the changes. I have already pushed the changes to a new branch here : https://github.com/facebookresearch/mmf/tree/lxmert_backup

eltoto1219:
My bad! I just added you as a collaborator on my branch, if there are still
permission errors, please let me know!

PR #343, opened 2020-06-18
[feat] Allow feature readers through registry, factory for readers

apsdehal:
Summary:
Allows adding new feature readers through registry.
Also, adds a factory for creating readers. Further, adds numpy feature reader
as a fallback reader in case none of the requirements are matched.

Differential Revision: D22108538


mmf/datasets/databases/readers/feature_readers.py
```diff
@@ -46,26 +50,23 @@ def _init_reader(self):
...
+        elif self._is_numpy:
+            self.feat_reader = NumpyFeatureReader()
         else:
             raise TypeError("unknown image feature format")
 
     def read(self, image_feat_path):
         if not image_feat_path.endswith("npy"):
             return None
+
+        self._is_numpy = True
```
vedanuj, line 69: Should this always be true?

mmf/datasets/databases/readers/feature_readers.py
```diff
@@ -79,28 +80,79 @@ def read(self, image_feat_path):
...
+        else:
+            feature_reader_name = feature_reader_config.get("type", "default")
+            feature_reader_params = feature_reader_config.get("params", {})
+            params.update(feature_reader_params)
+            feature_reader = cls.create(feature_reader_name, **params)
+
+        return feature_reader
+
+
+class FasterRCNNFeatureReader(MMFFeatureReader):
```
vedanuj, line 120: Can we add all these classes to registry so we can use them by specifying in configs going forward?

PR #345, opened 2020-06-19
[feat] Add MoViE+MCAN model

vedanuj:
- Adding the MoVie+MCAN model (VQA 2020 Challenge Winner)

TODOs:

- [x] Add typings to movie_mcan.py
- [x] Upload the annotations to zoo
- [x] Upload best model to zoo
- [x] Cleanup feature_readers.py
- [x] Update Readme


mmf/datasets/databases/readers/feature_readers.py
```diff
@@ -64,15 +64,15 @@ def _init_reader(self):
             raise TypeError("unknown image feature format")
```
vedanuj, line 64: Changed in this file is temporary until we upload features in proper format

mmf/modules/layers.py
```diff
@@ -665,3 +670,84 @@ def forward_all(self, v, q, v_mask=True):
...
+    """
+
+    def __init__(self, img_dim: int, ques_dim: int):
+        super().__init__()
+        self.out_dim = img_dim * 2
+        self.linear_cga = nn.ModuleList(
+            [nn.Linear(img_dim, self.out_dim) for _ in range(2)]
+        )
+        self.linear_cbn = nn.ModuleList(
+            [nn.Linear(1024, self.out_dim) for _ in range(2)]
```
kienduynguyen, line 700: Change 1024 to img_dim

kienduynguyen, line 700: Change 2014 to img_dim

mmf/modules/attention.py
```diff
@@ -156,3 +160,107 @@ def forward(self, image_feat, question_embedding, image_locs=None):
                 masked_attention = self._mask_attentions(attention, image_locs)
 
         return masked_attention
+
+
+class MultiHeadAttention(nn.Module):
```
apsdehal, line 165: Use torch's multiheadattention implementation directly?

mmf/modules/attention.py
```diff
@@ -156,3 +160,107 @@ def forward(self, image_feat, question_embedding, image_locs=None):
...
+        k = self.linears[1](k).view(b, -1, self.h, self.d_k).transpose(1, 2)
+        v = self.linears[2](v).view(b, -1, self.h, self.d_k).transpose(1, 2)
+
+        x, self.p_attn = self.qkv_attention(q, k, v, mask=mask, dropout=self.dropout)
+        x = x.transpose(1, 2).contiguous().view(b, -1, self.h * self.d_k)
+
+        return self.linears[-1](x)
+
+
+class SelfAttention(nn.Module):
```
apsdehal, line 212: Same here, these are implemented for Transformer module in torch. Maybe have a look?

mmf/modules/bottleneck.py
```diff
@@ -0,0 +1,252 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+from typing import Optional, Tuple, Type
+
+import torch
+import torch.nn as nn
+
+
+def conv3x3(
```
apsdehal, line 9: Use torchvision implementation directly? https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L24

mmf/modules/bottleneck.py
```diff
@@ -0,0 +1,252 @@
...
+        kernel_size=3,
+        stride=stride,
+        padding=dilation,
+        groups=groups,
+        bias=False,
+        dilation=dilation,
+    )
+
+
+def conv1x1(in_planes: int, out_planes: int, stride: int = 1) -> torch.Tensor:
```
apsdehal, line 25: https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L30

mmf/modules/bottleneck.py
```diff
@@ -0,0 +1,252 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
apsdehal, line 1: A lot of code in this file is already in torchvision. Let's use that and remove code duplication.

mmf/modules/layers.py
```diff
@@ -665,3 +670,84 @@ def forward_all(self, v, q, v_mask=True):
         p = nn.functional.softmax(expanded_logits, 2)
 
         return p.view(-1, self.glimpse, v_num, q_num), logits
+
+
+class TripleLinear(nn.Module):
```
apsdehal, line 675: Docstring?

mmf/modules/optimizers.py
```diff
@@ -1,7 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
+from torch.optim import Adam
 from transformers.optimization import AdamW
 
 from mmf.common.registry import registry
 
+registry.register_optimizer("adam")(Adam)
```
apsdehal, line 8: This can be directly used. Why register?

mmf/modules/bottleneck.py
kienduynguyen: Only conv1x1 and conv3x3 are in torchvision. The MoVie bottleneck is different from the one in torchvision since we have the condition vector and output 2 tensors instead of 1

mmf/modules/optimizers.py
kienduynguyen: @vedanuj We can delete it since I also train the network using AdamW

mmf/modules/attention.py
vedanuj: We will replace this in next PRs as this changes our model weigths from the challenge. Same for other methods from torch transformer.

mmf/configs/zoo/datasets.yaml
```diff
@@ -264,6 +264,14 @@ vqa2:
       - url: mmf://datasets/vqa2/defaults/extras.tar.gz
         file_name: extras.tar.gz
         hashcode: e45c3dbef61e5cc550b08ed7299c88cf7937efa219d071342b8ed7a6a5642a55
+  grid:
```
apsdehal, line 267: A comment here would be helpful. Do you want to really use `grid`? Seems too generic.

mmf/datasets/databases/readers/feature_readers.py
```diff
@@ -64,15 +64,18 @@ def _init_reader(self):
...
 
         if self.feat_reader is None:
             # Currently all lmdb features are with ndim == 2 so we are
             # avoiding loading the lmdb to determine feature ndim
             if not self.base_path.endswith(".lmdb") and self.ndim is None:
-                feat = np.load(image_feat_path)
+                if image_feat_path.endswith("npy"):
+                    feat = np.load(image_feat_path)
+                elif image_feat_path.endswith("pth"):
+                    feat = torch.load(image_feat_path)
```
apsdehal, line 78: NIT: I would also map the storage to cpu? In case the tensor saved was on GPU

apsdehal: I see that you did that below, might update this one as well?

mmf/models/movie_mcan.py
```diff
@@ -0,0 +1,270 @@
...
+    TextEmbedding,
+    TwoBranchEmbedding,
+)
+from mmf.modules.encoders import ImageFeatureEncoder
+from mmf.modules.layers import BranchCombineLayer, ClassifierLayer
+from mmf.utils.general import filter_grads
+
+
+@registry.register_model("movie_mcan")
+class MoVieMcan(BaseModel):
```
apsdehal, line 22: Something to keep in mind but not important as of now for speed:

Since the code has a lot of duplication w.r.t. to Pythia class, I would just take out the common parts and create a fusion mixin. And then use that mixin both in Pythia and here to reuse common parts. then, it also won't look like that MovieMCAN is built on top of Pythia.

mmf/modules/bottleneck.py
```diff
@@ -0,0 +1,228 @@
...
+            nn.Sigmoid(),
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = x * self.se(x)
+
+        return x * self.attn(x)
+
+
+class FrozenBatchNorm2d(nn.Module):
```
apsdehal, line 52: Please see https://github.com/pytorch/vision/blob/39021408587eb252ebb842a54195d840d6d76095/torchvision/ops/misc.py#L43

apsdehal: Might make sense to inherit it and call it `MovieMCANFrozenBatchNorm2d`

mmf/configs/zoo/datasets.yaml
vedanuj: Sure I will add a comment.

PR #349, opened 2020-06-22
[feat] Add GQA dataset

vedanuj:
- Adds the GQA dataset and Masked GQA dataset

Required for #339 LXMERT model pretraining and downstream task

Test Plan:
- Test for VisualBERT model

### DIFF
--- a/mmf/configs/datasets/gqa/defaults.yaml
+++ b/mmf/configs/datasets/gqa/defaults.yaml
@@ -1,0 +1,62 @@
+dataset_config:
+  gqa:
+    data_dir: ${env.data_dir}/datasets
+    depth_first: false
+    fast_read: false
+    use_images: false
+    use_features: true
+    zoo_requirements:
+    - gqa.defaults
+    features:
+      train:
+      - gqa/defaults/features/gqa.lmdb
+      val:
+      - gqa/defaults/features/gqa.lmdb
+      test:
+      - gqa/defaults/features/gqa.lmdb
+    annotations:
+      train:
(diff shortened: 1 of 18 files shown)


mmf/datasets/builders/gqa/masked_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+    def __getitem__(self, idx):
+        sample_info = self.annotation_db[idx]
+        current_sample = Sample()
+
+        if self._use_features is True:
+            features = self.features_db[idx]
+            current_sample.update(features)
+            image_labels = []
+
+            for i in range(features["image_feature_0"].shape[0]):
```
apsdehal, line 25: Shouldn't this only be done if the option 'use_image_feature_masks' is enabled? Otherwise, this is unnecessary calculation all of the time. I also think for long term maintenance purpose, we should convert this to a processor.

vedanuj:
We are not adding Scene graph support in this PR. We can add it later when requirement arises.

mmf/configs/datasets/gqa/defaults.yaml
```diff
@@ -0,0 +1,62 @@
...
+    data_dir: ${env.data_dir}/datasets
+    depth_first: false
+    fast_read: false
+    use_images: false
+    use_features: true
+    zoo_requirements:
+    - gqa.defaults
+    features:
+      train:
+      - gqa/defaults/features/gqa.lmdb
```
apsdehal, line 12: Would it make sense to be explicit about which feature type this is?

PR #367, opened 2020-06-27
[feature, fix] added support for update_frequency (v2)

eundersander:
This PR adds support for config.training.update_frequency:
```
# Run update_frequency=K batches with batch_size=N, accumulate gradients, then do 
# one update (one optimizer step). The effect is a large effective batch size of 
# KxN (without incurring the memory overhead of setting batch_size to KxN).
update_frequency: 1
```
https://www.internalfb.com/intern/tasks/?t=66569918

This is a revised approach; see also earlier PR https://github.com/facebookresearch/mmf/pull/292.

With this change, we have to be more clear with naming. Amanpreet and I agreed on this:
- we have no particular term for a single forward/backward pass for a single batch
- "update" == "iteration" == one optimizer step (will be more than one batch when update_frequency > 1)

The update logic in `BaseTrainer._backward` has mostly been moved to `_start_update` and `_finish_update`.

There's also a fix for `MultiDatasetLoader.__len__`. The existing bug is only present for the case that the train set size is not a multiple of per-worker batch size, which is probably rare in practice.

Test Plan:
I did ad-hoc testing of BaseTrainer.train() (`mmf_run config="projects/hateful_memes/configs/late_fusion/defaults.yaml" model="late_fusion" dataset=hateful_memes`) with print statements and varying `[batch_size, update_frequency, train set size]`, including end-of-epoch and completion of the train run.

### DIFF
--- a/mmf/utils/build.py
+++ b/mmf/utils/build.py
@@ -151,6 +151,7 @@ def build_dataloader_and_sampler(
             dataset_instance.dataset_name, dataset_instance.dataset_type
         ),
         num_workers=num_workers,
+        drop_last=False,  # see also MultiDatasetLoader.__len__
         **other_args,
     )
 
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -42,6 +42,10 @@ training:
     # Size of each batch. If distributed or data_parallel
     # is used, this will be divided equally among GPUs
     batch_size: 512
+    # Run update_frequency=K batches with batch_size=N, accumulate gradients, then do
+    # one update (one optimizer step). The effect is a large effective batch size of
+    # KxN (without incurring the memory overhead of setting batch_size to KxN).
+    update_frequency: 1
     # Number of workers to be used in dataloaders
     num_workers: 4
     # Some datasets allow fast reading by loading everything in the memory
--- a/mmf/datasets/multi_dataset_loader.py
+++ b/mmf/datasets/multi_dataset_loader.py
@@ -183,7 +183,10 @@ class MultiDatasetLoader:
 
     def __len__(self):
         # Since, this is iterator, we need to return total length == number of batches
-        return self._total_length // get_batch_size()
+        batch_size = get_batch_size()
+        # This assumes drop_last=False for all loaders. See also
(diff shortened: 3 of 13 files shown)


mmf/trainers/base_trainer.py
```diff
@@ -242,15 +242,45 @@ def train(self):
...
+
+                self._start_update()
+
+                for _ in range(num_batches_for_this_update):
+                    batch = next(batch_iter)
+                    self.profile("Batch load time")
+
+                    report = self._forward_pass(batch)
+                    loss = self._extract_loss(report)
+                    loss /= num_batches_for_this_update
```
eundersander, line 263: See also https://github.com/PyTorchLightning/pytorch-lightning/blob/82a20296e308a67c8d9202e4cbdf92a44b90b077/pytorch_lightning/trainer/training_loop.py#L590

mmf/trainers/base_trainer.py
```diff
@@ -242,15 +242,45 @@ def train(self):
...
+                    self._backward(loss)
+
+                    # accumulate necessary params for metric calculation
+                    if combined_report is None:
```
eundersander, line 267: See also https://github.com/facebookresearch/mmf/blob/45742b7dded3d38ed0109d7e107b9e7ae3dade66/mmf/trainers/base_trainer.py#L448

mmf/utils/build.py
```diff
@@ -140,6 +140,7 @@ def build_dataloader_and_sampler(
             dataset_instance.dataset_name, dataset_instance.dataset_type
         ),
         num_workers=num_workers,
+        drop_last=False,  # see also MultiDatasetLoader.__len__
```
eundersander, line 143: `drop_last=False` is the default so I'm not changing behavior here. I'm just making our choice explicit.

mmf/trainers/base_trainer.py
```diff
@@ -537,3 +573,11 @@ def predict(self, dataset_type):
 
             self.writer.write("Finished predicting")
             self.model.train()
+
+    def assert_iterator_finished(self, iter):
```
apsdehal, line 577: Probably move to this to util helper function.

PR #374, opened 2020-06-30
[fix] Reloads in spawned processes shouldn't setup imports again

### LINKED ISSUE #355: A code change affects other runs.
## 🐛 Bug

A code change affects other runs when num_workers > 0. 
If I change the code, then it seems a running file reimports python libraries. The imported python libraries are not fixed.
You can simply test by the following steps.

First, run any command.
Then make a code change; add "print("blahblah") to the text processor (e.g., BertTokenizer).
At some point (right before validation or between training iterations), the running file will output "blahblah".

## To Reproduce

1. Any command. I used Hateful memes dataset and Visual Bert.
2.
3.

## Expected behavior
If I make any changes, it should not affect other running files.

## Environment

Please copy and paste the output from the
environment collection script from PyTorch
(or fill out the checklist below manually).

You can run the script with:
```
# For security purposes, please check the contents of collect_env.py before running it.
python -m torch.utils.collect_env
```

 - PyTorch Version (e.g., 1.0):
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version:
 - GPU models and configuration:
 - Any other relevant information:

apsdehal:
- Fixes #355 partially
- In general, only setup_imports in the main processor
- This means that the child processes are using only shallow registry
which is fine as setup imports can cause issues in file not related with
syntax error
- Overall, this can't be fixed completely as by the nature of the spawn,
the child will still reload all of the code and if a syntax error is
there it will still break the worker. General suggestion is to keep two
copies of the code, one for development and one for running and keep
them in sync via GitHub

Test Plan:

Tested with command and reproduction steps provided by Ronghang

### DIFF
--- a/mmf/common/registry.py
+++ b/mmf/common/registry.py
@@ -406,4 +406,10 @@ class Registry:
 
 
 registry = Registry()
-setup_imports()
+
+# Only setup imports in main process, this means registry won't be
+# fully available in spawned child processes (such as dataloader processes)
+# but instantiated. This is to prevent issues such as
+# https://github.com/facebookresearch/mmf/issues/355
+if __name__ == "__main__":
+    setup_imports()


apsdehal:
Thanks for the suggestion. I added the suggestion to the issue linked. I had also looked into the issues you reported from PyTorch and PL. Those solutions felt somewhat clunky to me, and not in the scope of what mmf should do. Though it is good that they are linked here now, and future users can follow them from here. I would make this a part of FAQ doc that we have planned along with suggestions from us.

apsdehal:
For the part on copying the code, it won't work out of the box as we more of a do `pip install -e .` rather than direct script based run which means the environment would still load the existing code. A solution to that with conda environments is in my mind, I can try mixing it with our slurm scripts in future.

PR #383, opened 2020-07-02
Fix bug in truncate and refactor convert to index for readability

suzyahyah:
Summary:
Fix a bug in `truncate_seq_pair` where if `tokens_b` is not `None`, the `max_length` should be `max_length-3` because there are 3 special tokens, [CLS] tokens_a [SEP] tokens_b [SEP].

Refactor `tokens` and `segment_ids` in the same sequential form as `lm_label_ids` for readability, does not change the code logic.

Differential Revision: D22344634


vedanuj:
Please run linter so the CircleCI tests pass

PR #397, opened 2020-07-10
[refactor] Add transformer bbox processor

vedanuj:
- Refactor `bbox` position processing for transformer based models
- Fix feature path in masked coco dataset

Test Plan:

- Test with VilBERT model on VQA2, Masked COCO

### DIFF
--- a/mmf/datasets/builders/coco/masked_dataset.py
+++ b/mmf/datasets/builders/coco/masked_dataset.py
@@ -20,6 +20,9 @@ class MaskedCOCODataset(COCODataset):
         if self._use_features is True:
             features = self.features_db[idx]
 
+            if self.config.get("use_image_bbox_position", False):
+                features["image_info_0"] = self.bbox_processor(features["image_info_0"])
+
             if self.config.get("use_image_feature_masks", False):
                 current_sample.update(
                     {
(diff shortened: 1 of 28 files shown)


apsdehal:
We should add test for this processor as well.

mmf/datasets/processors/processors.py
```diff
@@ -1385,4 +1385,28 @@ def __call__(self, item):
             else:
                 # no masking token (will be ignored by loss function later)
                 image_labels.append(-1)
-        return image_labels
+        return torch.tensor(image_labels, dtype=torch.long)
+
+
+@registry.register_processor("transformer_bbox")
+class TransformerBboxProcessor(BaseProcessor):
```
apsdehal, line 1392: Can we add class documentation on what exactly this processor does?

mmf/datasets/processors/processors.py
```diff
@@ -1385,4 +1385,28 @@ def __call__(self, item):
...
+    def __init__(self, config, *args, **kwargs):
+        super().__init__(config, *args, **kwargs)
+
+    def __call__(self, item):
+        bbox = item["bbox"]
```
apsdehal, line 1397: This can be made parameters for processor config. Including `bbox_key`, `image_width` and `image_height`.

mmf/datasets/builders/coco/masked_dataset.py
```diff
@@ -20,6 +20,9 @@ def load_item(self, idx):
         if self._use_features is True:
             features = self.features_db[idx]
 
+            if self.config.get("use_image_bbox_position", False):
```
apsdehal, line 23: I find it little conflicting that we have two options doing same thing. If bbox processor is defined then it means we want to use it, no? Do we really need extra option?

PR #398, opened 2020-07-10
[feat] Add Base transformer class for modular multimodal transformers

vedanuj:
- Adds a base transformer model class with modular components for constructing a transformer based multi modal model. The modularity is divided as : 
```
        self.build_encoders()
        self.build_embeddings()
        self.build_transformer()
        self.build_heads()
        self.build_losses()
```
- Base model automatically loads an encoder specified by `transformer_base` in config. For example it can be `bert-base-uncased`, `roberta-base` etc.
- Ability to preprocess the input sample_list in `preprocess_sample ` according to model's requirement and also postprocess the output from the transformer according to requirement. For pretraining, we can add cls head and calculate loss on the output in the  `postprocess_output ` method. For classification we can calculate scores etc.

Test Plan:
- Tested with creating a new model using the base class and running on Hateful Memes, VQA2

### DIFF
--- a/mmf/models/transformers/base.py
+++ b/mmf/models/transformers/base.py
@@ -1,0 +1,222 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import os
+from dataclasses import dataclass
+from typing import Any, Dict, List, Type
+
+from omegaconf import OmegaConf
+from torch import Tensor, nn
+from transformers import AutoConfig, AutoModel
+
+from mmf.models import BaseModel
+from mmf.utils.configuration import get_mmf_cache_dir
+from mmf.utils.modeling import get_optimizer_parameters_for_bert
+
+
+@dataclass
+class BaseTransformerInput:
+    input_ids: Dict[str, Tensor]  # dict of input ids for all modalities
+    position_ids: Dict[str, Tensor]  # dict of position ids for all modalities
+    segment_ids: Dict[str, Tensor]  # dict of segment/token type ids for all modalities
+    masks: Dict[str, Tensor]  # dict of masks for all modalities
+
+
+@dataclass
+class BaseTransformerConfigType:
+    transformer_base: str  # name of transformer base model
+    training_head_type: str  # training head type used for initializing head
+    modalities: List[str]  # list of modalities for the model input
+    initializer_range: float  # std dev of the normal distribution to initialize layers
+    initializer_mean: float  # mean of the normal distribution to initialize layers
+    layer_norm_weight_fill: float  # layer norm weight initialization
+    random_initialize: bool  # random initialize whole network
+    freeze_base: bool  # freeze the base transformer
+    finetune_lr_multiplier: float  # finetune lr multiplier for base transformer
+
+
+class BaseTransformer(BaseModel):
+    def __init__(self, config: BaseTransformerConfigType):
+        """Initialize the config which is the model configuration and transformer_config
+        which is the config for the `transformer` base model.
+        """
+        super().__init__(config)
+        self.config = config
+        self.transformer_config = AutoConfig.from_pretrained(
+            config.transformer_base, **OmegaConf.to_container(config)
+        )
+
+    def build(self):
+        """Build the different parts of the multimodal transformer model and
+        initializes weights.
+        """
+        self.build_transformer()
+        self.build_encoders()
+        self.build_embeddings()
+        self.build_heads()
+        self.build_losses()
+
+        self.init_weights()
+
+    def get_optimizer_parameters(self, config: BaseTransformerConfigType):
+        return get_optimizer_parameters_for_bert(self, config)
+
+    def build_encoders(self):
+        """Build any encoders for different input modalities. Encoders are used while
+        preprocessing a sample. We the visual_encoder by default for raw image input.
+
+        Example ::
+
+            # For image
+            self.image_encoder = ImageEncoder(self.config)
+
+        """
+        return
+
+    def build_embeddings(self):
+        """Build the embeddings for the different input modalities.
+
+        Example ::
+
+            # For text
+            self.word_embeddings = nn.Embedding(
+                config.vocab_size, config.hidden_size, padding_idx=0
+            )
+            self.position_embeddings = nn.Embedding(
+                config.max_position_embeddings, config.hidden_size
+            )
+
+            # For image
+            self.img_embeddings = nn.Sequential(
+                nn.Linear(img_dim, config.hidden_size),
+                torch.nn.LayerNorm(config.hidden_size, eps=1e-12),
+            )
+        """
+        return
+
+    def build_transformer(self):
+        """Build the transformer encoder. This uses transformers AutoModel to load a
+        pretrained model given the name of the transformer based model. All the layers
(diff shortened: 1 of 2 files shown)


mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+        """Build any encoders for different input modalities. Encoders are used while
+        preprocessing a sample. We the visual_encoder by default for raw image input.
+
+        Example ::
+
+            # For image
+            self.vision_encoder = ImageEncoder(self.config)
+
+        """
+        pass
```
apsdehal, line 50: Instead of pass, return here.

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+from transformers import AutoConfig, AutoModel
+
+from mmf.common.typings import DictConfig
+from mmf.models import BaseModel
+from mmf.utils.configuration import get_mmf_cache_dir
+from mmf.utils.modeling import get_optimizer_parameters_for_bert
+
+
+class BaseTransformer(BaseModel):
+    def __init__(self, config: DictConfig):
```
apsdehal, line 18: We should document the config object here.

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+        pass
+
+    def _init_weights(self, module):
+        """Initialize the weights.
+        """
+        if isinstance(module, (nn.Linear, nn.Embedding)):
+            module.weight.data.normal_(mean=0.0, std=self.config.initializer_range)
+        elif isinstance(module, nn.module.LayerNorm):
+            module.bias.data.zero_()
+            module.weight.data.fill_(1.0)
```
apsdehal, line 131: Let's parameterize these as well.

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+        # Tie weights if required
+        self.tie_weights()
+
+    def preprocess_sample(self, sample_list: Dict[str, Any]):
+        """Preprocess the sample_list. Generate image input ids, image position ids,
+        image mask. Create a attention mask for all the masks for different modalities.
+        """
+        # TODO(vedanuj): Make this method simpler
+        # Image feature(img_input_ids) and position(img_pos_ids) tokens
+        if "image_feature_0" in sample_list:
```
apsdehal, line 156: Let's move this to a separate function for now so that its is easy to refactor later.

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+            sample_list.img_input_ids = sample_list.image_feature_0
```
apsdehal, line 157: Call it full `image_input_ids`?

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+            sample_list.img_pos_ids = sample_list.image_info_0.bbox
+        else:
+            sample_list.img_input_ids = self.vision_encoder(sample_list.image)
+            sample_list.img_pos_ids = sample_list.img_input_ids.new_tensor(
```
apsdehal, line 161:
```suggestion
            sample_list.image_position_ids = sample_list.image_input_ids.new_tensor(
```

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+                .unsqueeze(-1)
+            )
+
+        # Image mask
+        sample_list.image_mask = torch.ones_like(
+            sample_list.img_input_ids[:, :, 0], dtype=torch.long
+        )
+
+        # Attention mask
+        sample_list.attention_mask = torch.cat(
```
apsdehal, line 176: Attention mask shouldn't be assigned to sample list and should be calculated on the fly in the implementation.

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+            (sample_list.input_mask, sample_list.image_mask), dim=-1
+        )
+        sample_list.masked_lm_labels = getattr(sample_list, "lm_label_ids", None)
```
apsdehal, line 179: I am somewhat not a fan of editing sample_list dynamically as it is never clear to end user where a particular key came from. Maybe address it at root level? Or what is the problem with using `lm_label_ids`?

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+                torch.arange(0, sample_list.img_input_ids.size(1), dtype=torch.long)
+                .unsqueeze(0)
+                .expand(
+                    sample_list.img_input_ids.size(0), sample_list.img_input_ids.size(1)
+                )
+                .unsqueeze(-1)
+            )
+
+        # Image mask
+        sample_list.image_mask = torch.ones_like(
```
apsdehal, line 171: What if the mask was already present?

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+        """Preprocess the sample_list. Generate image input ids, image position ids,
+        image mask. Create a attention mask for all the masks for different modalities.
+        """
+        # TODO(vedanuj): Make this method simpler
+        # Image feature(img_input_ids) and position(img_pos_ids) tokens
+        if "image_feature_0" in sample_list:
+            sample_list.img_input_ids = sample_list.image_feature_0
+            sample_list.img_pos_ids = sample_list.image_info_0.bbox
+        else:
+            sample_list.img_input_ids = self.vision_encoder(sample_list.image)
```
apsdehal, line 160: `vision_encoder` seems weird, maybe call it `image_encoder`?

mmf/models/transformers/base.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+            new_lm_labels = torch.ones_like(sample_list.attention_mask) * -1
+            size_masked_lm_labels = sample_list.masked_lm_labels.size()
+            assert len(size_masked_lm_labels) == 2
+            new_lm_labels[
+                : size_masked_lm_labels[0], : size_masked_lm_labels[1]
+            ] = sample_list.masked_lm_labels
+            sample_list.masked_lm_labels = new_lm_labels
+
+    def forward(self, sample_list: Dict[str, Any]):
```
apsdehal, line 192: Document how would a SampleList in this case look like?

mmf/models/transformers/base.py
apsdehal: We should address this non-generic part in a separate PR as high pri.

mmf/models/transformers/base.py
vedanuj: Added dataclass for the config

mmf/models/transformers/base.py
vedanuj: This is removed from this base class for now

mmf/models/transformers/base.py
vedanuj: Base class now holds no opinions how the sample list should look like.

Muennighoff:
Should ALBERT also be possible with this new merge? Would I need to retrain e.g. VisualBERT from scratch on COCO or can I still use the pretrained versions if I change the transformer to e.g. roberta-base / albert etc?

PR #422, opened 2020-07-20
[feat] Generalized MM transformer

vedanuj:
Introduces a generalized version of multimodal transformer.

- Extends `BaseTransformer` and builds a modular multimodal transformer model
- Base transformer model can be any model from [transformers](https://github.com/huggingface/transformers)
- Supports n number of modalities
- Adds ability to add separate input, position and segment embeddings for each modality
- text modalities share weight with base transformer model embeddings
- position embeddings use separate embeddings for each modality
- token type/segment embeddings are shared/mean of base transformer token/segment embeddings (if available)

Test:
- Tested with HM, VQA2, OKVQA

### DIFF
--- a/mmf/configs/models/mmf_transformer/defaults.yaml
+++ b/mmf/configs/models/mmf_transformer/defaults.yaml
@@ -1,0 +1,31 @@
+model_config:
+  mmf_transformer:
+    transformer_base: bert-base-uncased
+    training_head_type: classification
+    num_labels: 2
+    modalities:
+      - type: text
+        key: text
+        segment_id: 0
+      - type: image
+        key: image
+        embedding_dim: 2048
+        position_dim: 1
+        segment_id: 1
+        layer_norm_eps: 1e-12
+        hidden_dropout_prob: 0.1
+    initializer_range: 0.02
+    initializer_mean: 0.0
+    token_noise_std: 0.01
+    token_noise_mean: 0.0
+    layer_norm_weight_fill: 1.0
+    random_initialize: false
+    freeze_transformer: false
+    freeze_image_encoder: false
+    finetune_lr_multiplier: 1
+    image_encoder:
+      type: resnet152
+      params:
+        pretrained: true
+        pool_type: avg
+        num_output_features: 1
--- a/mmf/models/mmf_transformer.py
+++ b/mmf/models/mmf_transformer.py
@@ -1,0 +1,327 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+from copy import deepcopy
+from typing import Any, Dict, Type
+
+import torch
+from mmf.common.registry import registry
+from mmf.common.typings import DictConfig
+from mmf.models.transformers.base import (
+    BaseTransformer,
+    BaseTransformerConfigType,
+    BaseTransformerInput,
+)
+from mmf.modules.encoders import MultiModalEncoderBase
+from torch import Tensor, nn
+from transformers.modeling_bert import BertPooler, BertPredictionHeadTransform
+
+
+class ImageEncoder(MultiModalEncoderBase):
+    """Extends the MultiModalEncoderBase class which builds the encoder based on
+    the config parameters. We can set the type of image encoder(resnet50, resnet152,
+    resnext50 etc) and other parameters like num of features, type of pooling etc.
+    """
+
+    def __init__(self, config: DictConfig):
+        super().__init__(config)
+
+    def build(self):
+        self.encoder = self._build_modal_encoder(self.config.image_encoder)
+
+    def forward(self, x: Tensor) -> Tensor:
+        return self.encoder(x)
+
+
+class MMFTransformerEmbeddings(nn.Module):
+    """Embedding class that can take any number of image or text modalities, each can
+    have their input id, position id and segment id. We generate embeddings of
+    dimension config.hidden_size for each and then first add the three embeddings
+    for each modality to have a modality specific embedding. We then concat the
+    modality specific embeddings to have a joint embedding.
+    """
+
+    def __init__(
+        self,
+        model_config: BaseTransformerConfigType,
+        transformer_config: Dict[str, Any],
+        transformer: Type[nn.Module],
+    ):
+        super().__init__()
+        self.model_config = model_config
+        self.transformer_config = transformer_config
+        self.build_layers()
+        self.init_weights(transformer)
+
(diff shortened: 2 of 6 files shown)


mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+class ImageEncoder(MultiModalEncoderBase):
+    """Extends the MultiModalEncoderBase class which builds the encoder based on
+    the config parameters. We can set the type of image encoder(resnet50, resnet152,
+    resnext50 etc) and other parameters like num of features, type of pooling etc.
+    """
+
+    def __init__(self, config: DictConfig):
+        super().__init__(config)
+
+    def build(self):
```
apsdehal, line 28: Do we need this extension? This isn't doing anything special on top of existing

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+        self.encoder = self._build_modal_encoder(self.config.image_encoder)
+
+    def forward(self, x: Tensor) -> Tensor:
```
apsdehal, line 31: I think this needs to be `Type[Tensor]` where Type is import from `typings`.

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+    have their input id, position id and segment id. We generate embeddings of
+    dimension config.hidden_size for each and then first add the three embeddings
+    for each modality to have a modality specific embedding. We then concat the
+    modality specific embeddings to have a joint embedding.
+    """
+
+    def __init__(
+        self,
+        model_config: Dict[str, Any],
+        transformer_config: Dict[str, Any],
```
apsdehal, line 46: Is this hugging face config?

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+        transformer: Type[nn.Module],
+    ):
+        super().__init__()
+        self.model_config = model_config
+        self.transformer_config = transformer_config
+        self._build_layers()
+        self._init_weights(transformer)
+
+    def _build_layers(self):
```
apsdehal, line 55: We should expose this as this won't fit all use cases.

mmf/configs/models/mmf_transformer/defaults.yaml
```diff
@@ -0,0 +1,22 @@
+model_config:
```
apsdehal, line 1: We should point to ConfigSchema definition here.

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+                self,
+                modality + "_dropout",
+                nn.Dropout(self.transformer_config.hidden_dropout_prob),
+            )
+
+        self.token_type_embeddings = nn.Embedding(
+            len(self.model_config.modalities), self.transformer_config.hidden_size
+        )
+
+    def _init_weights(self, transformer: Type[nn.Module]):
```
apsdehal, line 106: It would make sense to expose this as well.

apsdehal (commented):
First set of comments. Will review rest of it some time later.

apsdehal (changes requested):

mmf/models/mmf_transformer.py
vedanuj: We only require the image encoder to be built here

mmf/models/mmf_transformer.py
vedanuj: No, `Tensor` type works as it is.

mmf/models/mmf_transformer.py
vedanuj: Yes, base transformer config. I would keep the name as is now.

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+    ):
+        super().__init__()
+        self.model_config = model_config
+        self.transformer_config = transformer_config
+        self._build_layers()
+        self._init_weights(transformer)
+
+    def _build_layers(self):
+        for modality in self.model_config.modalities:
+            if "text" in modality:
```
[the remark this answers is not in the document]
vedanuj: Modified in latest version

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+                    )
+                    .unsqueeze(0)
+                    .expand(input_ids[modality].size()[:2])
+                )
+            total_embedding += getattr(self, modality + "_pos_embedding")(
+                position_ids[modality]
+            )
+
+            if modality not in segment_ids:
+                segment_ids[modality] = torch.zeros(
```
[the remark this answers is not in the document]
vedanuj: Now we ask segment_id in the config, and not take any assumptions. If not defined in config, we just skip it.

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+                            self.transformer_config.hidden_size, eps=1e-12
+                        ),
+                    ),
+                )
+            # Set the position embeddings
+            setattr(
+                self,
+                modality + "_pos_embedding",
+                nn.Embedding(
+                    self.transformer_config.max_position_embeddings,
```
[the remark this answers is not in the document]
vedanuj: Added this to config

mmf/models/mmf_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+
+        # Token Type Embeddings
+        if hasattr(transformer.embeddings, "token_type_embeddings"):
+            token_vocab_size = self.transformer_config.type_vocab_size
+            self.token_type_embeddings.weight.data[:token_vocab_size].copy_(
+                transformer.embeddings.token_type_embeddings.weight
+            )
+            for idx in range(token_vocab_size, len(self.model_config.modalities)):
+                self.token_type_embeddings.weight.data[idx].copy_(
+                    transformer.embeddings.token_type_embeddings.weight.data.mean(dim=0)
```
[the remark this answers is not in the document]
vedanuj: Add normal random noise

mmf/configs/models/mmf_transformer/defaults.yaml
vedanuj: Added to base transformer class

apsdehal (approved):

PR #436, opened 2020-07-29
[feat,fix] Logger overhaul, shift to Python logger

apsdehal:
- All logging uses python based logging after this
- Colored logging support inspired from d2
- Wherever touched moved to format strings
- Adds options for colored logging
- Moves torch version and cuda devices to top
- Moves all print and writer statements to logger
- Also, fixes the long failing windows tests for logger

What are the benefits of doing this?
- All libraries which use python logging will receives our logs now and
we receive theirs
- Proper division between stderr and stdout

Test Plan:

Tested locally

Reviewer suggestions:
This PR needs to be reviewed somewhat carefully to make sure I don't miss anything and as it touches all of the MMF.

### DIFF
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -33,8 +33,8 @@ training:
     # Whether MMF should log or not, Default: False, which means
     # mmf will log by default
     should_not_log: false
-    # Whether to capture stdout logs by the logger. Default: True
-    stdout_capture: true
+    # Whether the colored logs should be used
+    colored_logs: true
 
     # Tensorboard control, by default tensorboard is disabled
     tensorboard: false
--- a/mmf/utils/build.py
+++ b/mmf/utils/build.py
@@ -63,7 +63,7 @@ def build_model(config):
     model_class = registry.get_model_class(model_name)
 
     if model_class is None:
-        registry.get("writer").write("No model registered for name: %s" % model_name)
+        raise RuntimeError(f"No model registered for name: {model_name}")
     model = model_class(config)
 
     if hasattr(model, "build"):
--- a/mmf/utils/logger.py
+++ b/mmf/utils/logger.py
@@ -1,174 +1,255 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 import collections
+import functools
 import json
 import logging
 import os
 import sys
-from typing import Type
+from typing import Any, Dict, Union
 
+from termcolor import colored
+
+from mmf.common.registry import registry
 from mmf.utils.configuration import get_mmf_env
-from mmf.utils.distributed import is_master
+from mmf.utils.distributed import get_rank, is_master
 from mmf.utils.file_io import PathManager
 from mmf.utils.timer import Timer
 
 
-class Logger:
-    def __init__(self, config, name=None):
-        self._logger = None
-        self._is_master = is_master()
+@functools.lru_cache()
+def setup_output_folder(folder_only: bool = False):
+    """Sets up and returns the output file where the logs will be placed
+    based on the configuration passed. Usually "save_dir/logs/log_<timestamp>.txt".
+    If env.log_dir is passed, logs will be directly saved in this folder.
 
-        self.timer = Timer()
-        self.config = config
-        self.save_dir = get_mmf_env(key="save_dir")
-        self.log_format = config.training.log_format
-        self.time_format = "%Y_%m_%dT%H_%M_%S"
-        self.log_filename = "train_"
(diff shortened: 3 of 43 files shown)


mmf/configs/defaults.yaml
```diff
@@ -33,8 +33,8 @@ training:
     # Whether MMF should log or not, Default: False, which means
     # mmf will log by default
     should_not_log: false
-    # Whether to capture stdout logs by the logger. Default: True
-    stdout_capture: true
```
vedanuj, line 37: Are we not capturing stdout logs anymore or not having an option to do so? Or did I miss something?

mmf/utils/build.py
```diff
@@ -63,7 +63,7 @@ def build_model(config):
     model_class = registry.get_model_class(model_name)
 
     if model_class is None:
-        registry.get("writer").write("No model registered for name: %s" % model_name)
+        raise RuntimeError(f"No model registered for name: {model_name}")
```
vedanuj, line 66: Nice!! 👍

mmf/utils/logger.py
```diff
@@ -1,174 +1,251 @@
...
+    distributed_rank = get_rank()
+    handlers = []
+
+    if distributed_rank == 0:
+        logger.setLevel(logging.INFO)
+        ch = logging.StreamHandler(stream=sys.stdout)
+        ch.setLevel(logging.INFO)
+        if color:
+            formatter = ColorfulFormatter(
+                colored("%(asctime)s | %(name)s: ", "green") + "%(message)s",
```
vedanuj, line 106: Missing `%(levelname)s`

mmf_cli/run.py
```diff
@@ -31,12 +35,17 @@ def main(configuration, init_distributed=False, predict=False):
...
     config = build_config(configuration)
 
-    # Logger should be registered after config is registered
-    registry.register("writer", Logger(config, name="mmf.train"))
+    setup_logger(
+        color=config.training.colored_logs, disable=config.training.should_not_log
+    )
+    logger = logging.getLogger("mmf_cli.run")
+    logger.info(f"Torch version: {torch.__version__}")
+    log_device_names()
```
vedanuj, line 46: Nice 👍

tests/utils/test_logger.py
```diff
@@ -28,23 +29,34 @@ def setUpClass(cls) -> None:
...
 
     @classmethod
     def tearDownClass(cls) -> None:
         # Cleanup temp working dir.
-        for handler in cls.writer._logger.handlers:
+        for handler in cls.writer.handlers:
             handler.close()
         if cls._tmpdir is not None:
             # set ignore_errors as Windows throw error due to Permission
             shutil.rmtree(cls._tmpdir, ignore_errors=True)
```
vedanuj, line 43: Not sure what you mean by `long failing windows logger tests` in your PR description. Those were already passing earlier. One issue was regarding permissions removing the logger folder. If that is fixed by these changes, then we can remove `ignore_errors=True` here.

mmf/configs/defaults.yaml
apsdehal: We are capturing stdout always. There is a should_not_log option that basically does the same thing as setting this to false.

mmf/utils/logger.py
apsdehal: Colorful Formatter has color for log levels. INFO is green, and when warning or error are logged they are appended on front in red color. Check the source for Colorful Formatter.

tests/utils/test_logger.py
apsdehal: The log files with ":" were some how still failing with OSError. I fixed them by changing ":" to "_".

mmf/configs/defaults.yaml
vedanuj: `stdout_capture` here means capturing stdout to log files. It is different from `should_not_log`.

mmf/utils/logger.py
```diff
@@ -1,174 +1,251 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 import collections
```
vedanuj, line 3: Will be good to add a comment here that most of this is adapted from Detectron2.

tests/utils/test_logger.py
vedanuj: Thanks. PR #352 did fix some log file names. Maybe it missed some.

mmf/configs/defaults.yaml
apsdehal: Yes, discussed offline. The option isn't useful anymore.

PR #437, opened 2020-07-30
[fix] Captioning predictions for bsz 1, remove opinions in testreporter

apsdehal:
- Use object_to_byte and reverse for saving image ids
- Remove opinions about shapes and gathering in the test reporter based
on keys. Generalized logic now.

### DIFF
--- a/mmf/common/test_reporter.py
+++ b/mmf/common/test_reporter.py
@@ -3,7 +3,6 @@ import csv
 import json
 import os
 
-import torch
 from torch.utils.data import DataLoader, Dataset
 from torch.utils.data.distributed import DistributedSampler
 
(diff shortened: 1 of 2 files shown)


mmf/common/test_reporter.py
```diff
@@ -170,16 +161,13 @@ def add_to_report(self, report, model):
...
-
         if key in report:
             num_dims = report[key].dim()
             if num_dims == 1:
                 report[key] = gather_tensor(report[key]).view(-1)
-
-            elif num_dims == 2:
-                _, enc_size = report[key].size()
+            elif num_dims >= 2:
+                enc_size = report[key].size(-1)
```
vedanuj, line 169: Taking  `.size(-1)` will only take the last dimension. So I think this won't work when `num_dims > 2`

```suggestion
                enc_size = *report[key].size()[1:]
```

apsdehal: Thanks. This has been resolved.

Issue #431, opened 2020-07-26
Overriding annotations


gantoniou92:
Hi MMF team,

I have a question regarding overriding annotations using dot list. It is suggested (https://mmf.sh/docs/notes/configuration#command-line-dot-list-override) that one can override annotations e.g. 

`dataset_config.hateful_memes.annotations.train[0]=x`. 

Doing something similar 

`dataset_config.hateful_memes.annotations.train[0]=hateful_memes/defaults/annotations/new_train.jsonl`, 

it doesn't seem that any overriding is happening, rather the original annotation file train.jsonl is loading. However, if I use 

`dataset_config.hateful_memes.annotations.train=hateful_memes/defaults/annotations/new_train.jsonl`

I get the message:

> Overriding option dataset_config.hateful_memes.annotations.train to hateful_memes/defaults/annotations/new_train.jsonl

thus probably loading the new annotation file. In addition, in this case the produced config file the corresponding part looks:

annotations:
      train: hateful_memes/defaults/annotations/new_train.jsonl
      val:
      - hateful_memes/defaults/annotations/dev.jsonl
      test:
      - hateful_memes/defaults/annotations/test.jsonl

Q: Is this expected behaviour or am I missing something in my understanding of this functionality in mmf?

Thanks in advance for your help.

apsdehal:
Fixed by #438 in master. Closing.

PR #438, opened 2020-07-30
[fix] Override of final array elements through dot element

charlesxbai:
Should check if we reach the end of the dot_list when overriding config with args. This should fix issue #431 

Test:
```
mmf_run config=projects/m4c/configs/textvqa/defaults.yaml \
    datasets=textvqa \
    model=m4c \
    run_type=train_val \
    training.lr_steps[1]=100 \
    dataset_config.textvqa.zoo_requirements[0]="test" \
```

Output:
```
Overriding option config to projects/m4c/configs/textvqa/defaults.yaml
Overriding option datasets to textvqa
Overriding option model to m4c
Overriding option run_type to train_val
Overriding option training.lr_steps[1] to 100
Overriding option dataset_config.textvqa.zoo_requirements[0] to test
```

Also: `pytest ./tests/utils/test_configuration.py`


charlesxbai:
Added test

tests/configs/test_config_overrides.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import unittest
+
+from mmf.utils.configuration import Configuration
+from mmf.utils.env import setup_imports
+from mmf.utils.general import get_mmf_root
+from tests.test_utils import dummy_args
+
+
+class TestConfigOverrides(unittest.TestCase):
```
apsdehal, line 11: These look good and ready to go. Can you move these to: https://github.com/facebookresearch/mmf/blob/master/tests/utils/test_configuration.py and then we can import it to phabricator.

PR #458, opened 2020-08-06
[fix] raise value error rather than exception; fixed formating with black

ytsheng:
[fix] raise value error rather than exception; fixed formating with black

- I apologize with the premature land, I kicked off transfering code before I saw the comments. 
- Test with ``pytest tests``
- This passes tests

Question, would you be open to using git precommit hooks to run isort, black automatically?


vedanuj:
You can use pre-commit hooks. We have the config in the repo.

PR #459, opened 2020-08-06
[fix] Support for reading raw COCO images for VQA2 and COCO

suzyahyah:
Summary:
VQA2 and COCO dataset readers `dataset.py` previously only support reading in from feature file.

After this PR, it should now be compatible with a model which reads in raw images. Should not affect any models which previously relied on `use_images: false`, `use_features: true`

Changes:

* Loads images in `{vqa2,coco}/{masked_,}dataset.py` if `self._use_features` flag is False

* Included image transform in `faim/mmf/mmf/datasets/builders/vqa2/dataset.py:def init_processors(self)`

* Enabled npy loading for internal
`faim/mmf/mmf/datasets/databases/annotation_database.py:def _load_npy()`

* Configs -- The main difference for configs is the inclusion of image_processor in this config file: `mmf/configs/datasets/vqa2/with_raw_images.yaml` and the dataset config `use_images` and `use_features` flags.

```
dataset_config:
  vqa2:
    use_images: true
    use_features: false
  images:
    train:
    - coco/defaults/images/train2014
    val:
    - coco/defaults/images/val2014
    test:
    - coco/defaults/images/test2015

Differential Revision: D22979686

### DIFF
--- a/mmf/configs/datasets/vqa2/with_raw_images.yaml
+++ b/mmf/configs/datasets/vqa2/with_raw_images.yaml
@@ -1,0 +1,75 @@
+dataset_config:
+  vqa2:
+      data_dir: ${env.data_dir}/datasets
+      depth_first: false
+      fast_read: false
+      use_images: true
+      use_features: false
+      zoo_requirements:
+      - coco.defaults
+      - vqa2.defaults
+      images:
+        train:
+        - coco/defaults/images/train2014
+        val:
+        - coco/defaults/images/val2014
+        test:
+        - coco/defaults/images/test2015
+      features:
+        train:
+        - coco/defaults/features/trainval2014.lmdb
+        val:
+        - coco/defaults/features/trainval2014.lmdb
+        test:
+        - coco/defaults/features/test2015.lmdb
+      annotations:
+        train:
+        - vqa2/defaults/annotations/imdb_train2014.npy
+        val:
+        - vqa2/defaults/annotations/imdb_val2014.npy
+        test:
+        - vqa2/defaults/annotations/imdb_test2015.npy
+      max_features: 100
+      processors:
+        answer_processor:
+          type: vqa_answer
+          params:
+            num_answers: 10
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: mmf/datasets/builders/coco/dataset.py [hunk 1]
```diff
@@ -38,9 +38,12 @@ def load_item(self, idx):
 
         current_sample.image_id = object_to_byte_tensor(sample_info["image_id"])
 
-        if self._use_features is True:
+        if self._use_features:
             features = self.features_db[idx]
             current_sample.update(features)
+        else:
+            image_path = str(sample_info["image_name"]) + ".jpg"
+            current_sample.image = self.image_db.from_path(image_path)["images"][0]
```
apsdehal, line 46: The transforms should be assigned to image_db so they are applied automatically.

apsdehal: I believe this needs to be fixed before this can land.

suzyahyah: Thanks, amended!

mmf/configs/datasets/vqa2/with_raw_images.yaml
```diff
@@ -0,0 +1,75 @@
+dataset_config:
```
apsdehal, line 1: There is lot of redundancy here which decreases readability. You would want to only override the stuff that needs to be overridden and skip common parts with defaults.

suzyahyah: when do we include defaults explicitly? I see this in some places and not others

```
includes:
- ./defaults.yaml
```

My guess is that we don't need to explicitly include it for mmf/configs/{models,datasets}/<x>/ but need to explicitly include it in mmf/projects/<x>/configs ?

apsdehal: The defaults defined in `config_path` attribute are included automatically. Everything else needs to be included separately. In this case, since `with_raw_images.yaml` is an extension, it will need to be included separately.

PR #460, opened 2020-08-06
[feat] Support for more than 2 segments in MMBT

apsdehal:
Summary:
Segment embedding other than first 2 dim are initialized from the average of first 2 embeddings

This is required for the case where there are 3 segments such as post, ocr, and features.

Differential Revision: D22970794

### DIFF
--- a/mmf/models/mmbt.py
+++ b/mmf/models/mmbt.py
@@ -351,6 +351,34 @@ class MMBTBase(MultiModalEncoderBase):
         if self.config.use_modal_end_token:
             modal_end_token = sample_list.input_ids[:, -1].clone().detach()
 
+        if hasattr(sample_list, "modal_token_type_ids"):
+            modal_token_type_ids = sample_list.modal_token_type_ids
+        else:
+            token_value = 0
+            segment_ids = sample_list.segment_ids
+            max_id = segment_ids.max()
+            min_id = segment_ids.min()
+            # Case of only one segment
+            if max_id == min_id:
+                # If max_id is greater than 0, that means text is at 0 segment
+                # which means modal will be at 1
+                # In other case, it will be zero, which it already is
+                if max_id == 0:
+                    token_value = 1
+            else:
+                max_segment = self.config.text_encoder.params.get("num_segments", 2) - 1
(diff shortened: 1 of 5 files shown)


mmf/models/mmbt.py
```diff
@@ -352,6 +352,43 @@ def forward(self, sample_list):
...
+                # text segments start from 0 which means modal will
+                # be last, otherwise, it is 0, which is already is
+                if max_id != max_segment:
+                    modal_token_type_ids = torch.full(
+                        (input_modal.size(0), 1),
+                        fill_value=max_segment,
+                        dtype=torch.long,
+                        device=input_modal.device,
+                    )
+
```
vedanuj, line 390: This can probably be refactored to first determine the proper value of the modal segment, and then set it using `torch.full` once (instead of using torch.zeros, torch.ones for different use cases).

```
modal_token_type_ids = torch.full(
                        (input_modal.size(0), 1),
                        fill_value=max_segment,
                        dtype=torch.long,
                        device=input_modal.device,
                    )
```

mmf/models/mmbt.py
```diff
@@ -352,6 +352,43 @@ def forward(self, sample_list):
         if self.config.use_modal_end_token:
             modal_end_token = sample_list.input_ids[:, -1].clone().detach()
 
+        if hasattr(sample_list, "modal_token_type_ids"):
+            modal_token_type_ids = sample_list.modal_token_type_ids
+        else:
+            modal_token_type_ids = torch.zeros(
+                (input_modal.size(0)),
```
vedanuj, line 359: Will the shape of the segment be always Bx1?

apsdehal: We are relying on the broadcasting here.

mmf/models/mmbt.py
apsdehal: Refactored. Thanks for the suggestion.

apsdehal:
Original implementation of MMBT had a bug in which both modal and text were using 0 as segment ids. I am retraining the models on HM and will update the PR with the updated models as well. The rest of the content in the PR is ready for review.

PR #461, opened 2020-08-06
[fix] Allow user to define 'text_b' for bert_tokenizer text processor.

suzyahyah:
Summary:
Previously `bert_tokenizer` text processor assumed there are no `text_b` tokens.

We still preserve this default logic (tokens_b defaults to None), but additionally allow the user to specify `text_b` if desired.

Differential Revision: D22988310

### DIFF
--- a/mmf/datasets/processors/bert_processors.py
+++ b/mmf/datasets/processors/bert_processors.py
@@ -145,6 +145,7 @@ class BertTokenizer(MaskedTokenProcessor):
         self._probability = 0
 
     def __call__(self, item):
+
         if "text" in item:
             text_a = item["text"]
         else:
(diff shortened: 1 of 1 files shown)


mmf/datasets/processors/bert_processors.py
```diff
@@ -141,17 +141,23 @@ def __call__(self, item):
 class BertTokenizer(MaskedTokenProcessor):
     def __init__(self, config, *args, **kwargs):
         super().__init__(config, *args, **kwargs)
-        self._probability = 0
```
apsdehal, line 144: This option is useful to explicitly inform user that probability is zero. This is for readability purposes.

vedanuj:
Thanks for the PR. There is one linter test failing. It might not be related to this change but previous change to the file `bert_processors.py`. Could you fix that and re export?

mmf/datasets/processors/bert_processors.py
```diff
@@ -168,6 +178,7 @@ class MultiSentenceBertTokenizer(BertTokenizer):
     bert tokenizer separately and indices will be reshaped as single
     tensor. Segment ids will also be increasing in number.
     """
+
```
apsdehal, line 181: Is this required change?

Issue #453, opened 2020-08-05
Movie+Mcan: grid feature size? (2048, 26, 19) vs (2048,25,19) 


CCYChongyanChen:
## ❓ Questions and Help
Hi, I am trying to run the grid+MCAN. 
I extracted the grid features following https://github.com/facebookresearch/grid-feats-vqa and stored the features in .pth. Each .pth has a size [1, 2048, 26,19].
When I run the code, I mean a RuntimeError: The expanded size of the tensor (25) must match the existing size (26) at non-singleton dimension 1.  Target sizes: [2048, 25, 19].  Tensor sizes: [2048, 26, 19]

The full traceback is attached.

Traceback (most recent call last):
  File "/home/cc67459/MMF2/bin/mmf_run", line 33, in <mo                     dule>
    sys.exit(load_entry_point('mmf', 'console_scripts',                      'mmf_run')())
  File "/home/cc67459/MMF2/mmf_8_2/mmf/mmf_cli/run.py",                      line 118, in run
    nprocs=config.distributed.world_size,
  File "/home/cc67459/MMF2/lib/python3.7/site-packages/t                     orch/multiprocessing/spawn.py", line 200, in spawn
    return start_processes(fn, args, nprocs, join, daemo                     n, start_method='spawn')
  File "/home/cc67459/MMF2/lib/python3.7/site-packages/t                     orch/multiprocessing/spawn.py", line 158, in start_proce                     sses
    while not context.join():
  File "/home/cc67459/MMF2/lib/python3.7/site-packages/t                     orch/multiprocessing/spawn.py", line 119, in join
    raise Exception(msg)
Exception:

-- Process 2 terminated with the following error:
Traceback (most recent call last):
  File "/home/cc67459/MMF2/lib/python3.7/site-packages/t                     orch/multiprocessing/spawn.py", line 20, in _wrap
    fn(i, *args)
  File "/home/cc67459/MMF2/mmf_8_2/mmf/mmf_cli/run.py",                      line 66, in distributed_main
    main(configuration, init_distributed=True, predict=p                     redict)
  File "/home/cc67459/MMF2/mmf_8_2/mmf/mmf_cli/run.py",                      line 56, in main
    trainer.train()
  File "/home/cc67459/MMF2/mmf_8_2/mmf/mmf/trainers/mmf_                     trainer.py", line 108, in train
    self.training_loop()
[...]

PR #464, opened 2020-08-09
[feat] Padded grid features

vedanuj:
Fixes issue #453 . 

- Add support for variable feature size for grid based features.
- This will enable using features from https://github.com/facebookresearch/grid-feats-vqa repository

Test Plan:

Tested Movie+Mcan model with variable sized features.

### DIFF
--- a/mmf/datasets/databases/readers/feature_readers.py
+++ b/mmf/datasets/databases/readers/feature_readers.py
@@ -57,7 +57,7 @@ class FeatureReader:
         elif self.ndim == 3 and not self.depth_first:
             self.feat_reader = Dim3FeatureReader()
         elif self.ndim == 4 and self.depth_first:
-            self.feat_reader = CHWFeatureReader()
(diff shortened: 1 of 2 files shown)


mmf/datasets/databases/readers/feature_readers.py
```diff
@@ -57,7 +57,7 @@ def _init_reader(self):
         elif self.ndim == 3 and not self.depth_first:
             self.feat_reader = Dim3FeatureReader()
         elif self.ndim == 4 and self.depth_first:
-            self.feat_reader = CHWFeatureReader()
+            self.feat_reader = CHWFeatureReader(self.max_features)
```
apsdehal, line 60: Would it be not possible to use PaddedFasterRCNN or refactor this logic of padding?

vedanuj: The features from `grid-feat-vqa` are of CHW format and doesn't have any `_info` files that are used in PaddedFasterRCNN. So I would keep those two separate.

PR #468, opened 2020-08-10
[feat] Open image features with PathManager

vedanuj:
Summary:
- Adds ability to read features from different URIs (Manifold, Everstore URIs) in addition to file paths.
- Adds target for lmdb conversion/extraction

Differential Revision: D23021466

### DIFF
--- a/mmf/datasets/databases/readers/feature_readers.py
+++ b/mmf/datasets/databases/readers/feature_readers.py
@@ -1,11 +1,27 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+
 import os
 import pickle
+from typing import Any
 
 import lmdb
 import numpy as np
 import torch
 from mmf.utils.file_io import PathManager
+
+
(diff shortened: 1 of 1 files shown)


mmf/datasets/databases/readers/feature_readers.py
```diff
@@ -90,25 +92,29 @@ def read(self, image_feat_path):
...
+            tmp = np.load(f)
         _, _, c_dim = tmp.shape
         image_feature = torch.from_numpy(np.reshape(tmp, (-1, c_dim)))
         return image_feature, None
 
 
 class HWCFeatureReader:
     def read(self, image_feat_path):
-        tmp = np.load(image_feat_path)
+        with PathManager.open(image_feat_path, "rb") as f:
```
apsdehal, line 116: Convert this to a private utility function in this file and use it to load the field. That would be much cleaner and make it easy to make any updates to this in future.

PR #469, opened 2020-08-11
[feat] Install glove only when the processor is called

Miiira:
Embeddings download happens in Vocab(..), which was previously called in __init__. We put it in __call__ instead to delay the download.

### DIFF
--- a/mmf/datasets/processors/processors.py
+++ b/mmf/datasets/processors/processors.py
@@ -374,9 +374,17 @@ class GloVeProcessor(VocabProcessor):
 
             vocab_processor_config.vocab.type = "pretrained"
 
-        super().__init__(vocab_processor_config, *args, **kwargs)
+        self._init_extras(vocab_processor_config)
+        self.config = vocab_processor_config
+        self._already_downloaded = False
+        self._args = args
+        self._kwargs = kwargs
 
     def __call__(self, item):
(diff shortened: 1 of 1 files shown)


mmf/datasets/processors/processors.py
```diff
@@ -374,9 +374,17 @@ def __init__(self, config, *args, **kwargs):
...
-        super().__init__(vocab_processor_config, *args, **kwargs)
+        self._init_extras(vocab_processor_config)
+        self.config = vocab_processor_config
+        self._already_downloaded = False
+        self._args = args
+        self._kwargs = kwargs
 
     def __call__(self, item):
+        if not self._already_downloaded:
+            self.vocab = Vocab(*self._args, **self.config.vocab, **self._kwargs)
```
apsdehal, line 385: Please make sure that this is only initialized in the main process.

Miiira: I think this is handled in Vocab(..):
```
if is_master():
    vocab.pretrained_aliases[embedding_name](cache=vector_cache)
synchronize()
```

apsdehal: Sounds good! Please follow the contributing guidelines for proper commit message. Then, edit the PR title and reimport to phabricator.

PR #470, opened 2020-08-11
[feat] added split dataset feature

ytsheng:
* allow user to specify x% of dataset to split from train, example yaml shown below:
```
  textvqa:
    zoo_requirements:
    - textvqa.defaults
    - textvqa.ocr_en
    split_train:
      val: 0.2
      test: 0.1
      seed: 42
    features:
      train:
      - textvqa/defaults/features/open_images/detectron.lmdb,textvqa/ocr_en/features/ocr_en_frcn_features.lmdb
   processors:
```
* To test: `pytests test`
* See logs, 24221 is 0.7 of 34602 which is the total size of the train after calculation, etc.
Screen Shot 2020-08-10 at 11 47 12 PM

### DIFF
--- a/mmf/datasets/mmf_dataset_builder.py
+++ b/mmf/datasets/mmf_dataset_builder.py
@@ -3,7 +3,6 @@ import copy
 import os
 import typing
 import warnings
-import numpy as np
 
 import numpy as np
 
@@ -13,6 +12,7 @@ from mmf.datasets.concat_dataset import MMFConcatDataset
 from mmf.datasets.subset_dataset import MMFSubset
 from mmf.utils.configuration import get_global_config, get_mmf_env, get_zoo_config
 from mmf.utils.general import get_absolute_path
+
 
 class MMFDatasetBuilder(BaseDatasetBuilder):
     ZOO_CONFIG_PATH = None
--- a/mmf/datasets/mmf_dataset_builder.py
+++ b/mmf/datasets/mmf_dataset_builder.py
@@ -1,14 +1,18 @@
 import collections
+import copy
 import os
 import typing
 import warnings
+import numpy as np
+
+import numpy as np
 
 import mmf.utils.download as download
 from mmf.datasets.base_dataset_builder import BaseDatasetBuilder
 from mmf.datasets.concat_dataset import MMFConcatDataset
+from mmf.datasets.subset_dataset import MMFSubset
 from mmf.utils.configuration import get_global_config, get_mmf_env, get_zoo_config
 from mmf.utils.general import get_absolute_path
-
 
 class MMFDatasetBuilder(BaseDatasetBuilder):
     ZOO_CONFIG_PATH = None
@@ -119,6 +123,49 @@ class MMFDatasetBuilder(BaseDatasetBuilder):
 
     def load(self, config, dataset_type, *args, **kwargs):
         self.config = config
+
+        new_config = config
+        split_dataset_from_train = self.config.split_train
+        if split_dataset_from_train:
+            split_range = self._calculate_split_for_dataset_type(dataset_type)
+            new_config = self._modify_dataset_config(config)
+
+        annotations = self._read_annotations(new_config, dataset_type)
+        if annotations is None:
+            return None
+
+        datasets = []
+        for imdb_idx in range(len(annotations)):
+            dataset_class = self.dataset_class
+            dataset = dataset_class(new_config, dataset_type, imdb_idx)
+            datasets.append(dataset)
+
+        dataset = MMFConcatDataset(datasets)
+        if split_dataset_from_train:
+            start, end = split_range
+            dataset_length = len(dataset)
+            start, end = round(start * dataset_length), round(end * dataset_length)
+            seed = new_config.split_train.seed
+            indices = np.random.RandomState(seed).permutation(dataset_length)[start:end]
+            dataset = MMFSubset(dataset, indices)
+
+        self.dataset = dataset
+        return self.dataset
+
+    def _modify_dataset_config(self, config):
+        modified_config = copy.deepcopy(config)
+        for data_type in config.split_train:
+            if data_type == "seed":
+                continue
+            if config.use_images:
+                modified_config.images[data_type] = config.images.train
+            if config.use_features:
+                modified_config.features[data_type] = config.features.train
+            modified_config.annotations[data_type] = modified_config.annotations.train
+
+        return modified_config
+
+    def _read_annotations(self, config, dataset_type):
         annotations = config.get("annotations", {}).get(dataset_type, [])
 
         # User can pass a single string as well
@@ -133,17 +180,25 @@ class MMFDatasetBuilder(BaseDatasetBuilder):
                 + "This dataset won't be used.".format(dataset_type)
             )
             return None
+        return annotations
 
-        datasets = []
+    def _calculate_split_for_dataset_type(self, dataset_type):
+        start = 0.0
+        for data_type in self.config.split_train:
+            if data_type == "seed":
+                continue
+            if dataset_type == data_type:
+                return (start, start + self.config.split_train[data_type])
+            start += self.config.split_train[data_type]
 
-        for imdb_idx in range(len(annotations)):
-            dataset_class = self.dataset_class
-            dataset = dataset_class(config, dataset_type, imdb_idx)
-            datasets.append(dataset)
+        if start >= 1.0:
+            raise ValueError(
+                "Ratios of val and test should not exceed 100%."
+                + "Need to leave some percentage for training."
+            )
 
-        dataset = MMFConcatDataset(datasets)
-        self.dataset = dataset
-        return self.dataset
+        if dataset_type == "train":
+            return (start, 1.0)
 
     def _download_based_on_attribute(
         self, resources, download_path, version, attribute
--- a/mmf/datasets/mmf_dataset_builder.py
+++ b/mmf/datasets/mmf_dataset_builder.py
@@ -1,17 +1,17 @@
 import collections
-import copy
 import os
 import typing
 import warnings
-
-import numpy as np
+from copy import deepcopy
 
 import mmf.utils.download as download
+import torch
 from mmf.datasets.base_dataset_builder import BaseDatasetBuilder
 from mmf.datasets.concat_dataset import MMFConcatDataset
 from mmf.datasets.subset_dataset import MMFSubset
 from mmf.utils.configuration import get_global_config, get_mmf_env, get_zoo_config
 from mmf.utils.general import get_absolute_path
+from omegaconf import open_dict
 
 
 class MMFDatasetBuilder(BaseDatasetBuilder):
@@ -24,7 +24,7 @@ class MMFDatasetBuilder(BaseDatasetBuilder):
         dataset_class=None,
         zoo_variation="defaults",
         *args,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(dataset_name)
         self.dataset_class = dataset_class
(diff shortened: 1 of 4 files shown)


mmf/datasets/subset_dataset.py
```diff
@@ -0,0 +1,15 @@
+from torch.utils.data.dataset import Subset
```
ytsheng, line 1: oh I need to add a comment on copyright... Will do.

ytsheng:
hold on the review for a moment, running into some git push issues.

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
 
     def load(self, config, dataset_type, *args, **kwargs):
         self.config = config
+
+        new_config = config
+        split_dataset_from_train = self.config.split_train
```
apsdehal, line 128: This will break existing datasets, use 
`self.config.get("split_train", False)`

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
...
+        self.dataset = dataset
+        return self.dataset
+
+    def _modify_dataset_config(self, config):
+        modified_config = copy.deepcopy(config)
+        for data_type in config.split_train:
+            if data_type == "seed":
+                continue
+            if config.use_images:
+                modified_config.images[data_type] = config.images.train
```
apsdehal, line 161: Use deepcopy to assign as well.

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
...
+            dataset_length = len(dataset)
+            start, end = round(start * dataset_length), round(end * dataset_length)
+            seed = new_config.split_train.seed
+            indices = np.random.RandomState(seed).permutation(dataset_length)[start:end]
+            dataset = MMFSubset(dataset, indices)
+
+        self.dataset = dataset
+        return self.dataset
+
+    def _modify_dataset_config(self, config):
```
apsdehal, line 155: Make this function's name more specific. `_modify_dataset_config_for_split`

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -133,17 +180,25 @@ def load(self, config, dataset_type, *args, **kwargs):
...
+            if data_type == "seed":
+                continue
+            if dataset_type == data_type:
+                return (start, start + self.config.split_train[data_type])
+            start += self.config.split_train[data_type]
+
+        if start >= 1.0:
+            raise ValueError(
+                "Ratios of val and test should not exceed 100%."
+                + "Need to leave some percentage for training."
```
apsdehal, line 197:
```suggestion
                + " Need to leave some percentage for training."
```

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -133,17 +180,25 @@ def load(self, config, dataset_type, *args, **kwargs):
...
+        start = 0.0
+        for data_type in self.config.split_train:
+            if data_type == "seed":
+                continue
+            if dataset_type == data_type:
+                return (start, start + self.config.split_train[data_type])
+            start += self.config.split_train[data_type]
+
+        if start >= 1.0:
+            raise ValueError(
```
apsdehal, line 195: This probably should be a warning in case of == 1.0 as I can launch this on val inference alone.

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
 
     def load(self, config, dataset_type, *args, **kwargs):
         self.config = config
+
+        new_config = config
+        split_dataset_from_train = self.config.split_train
+        if split_dataset_from_train:
+            split_range = self._calculate_split_for_dataset_type(dataset_type)
+            new_config = self._modify_dataset_config(config)
```
apsdehal, line 131: Why not directly assign to config?

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
...
+            datasets.append(dataset)
+
+        dataset = MMFConcatDataset(datasets)
+        if split_dataset_from_train:
+            start, end = split_range
+            dataset_length = len(dataset)
+            start, end = round(start * dataset_length), round(end * dataset_length)
+            seed = new_config.split_train.seed
+            indices = np.random.RandomState(seed).permutation(dataset_length)[start:end]
+            dataset = MMFSubset(dataset, indices)
```
apsdehal, line 150: We can probably move this to a separate function at this point

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
...
+            dataset = dataset_class(new_config, dataset_type, imdb_idx)
+            datasets.append(dataset)
+
+        dataset = MMFConcatDataset(datasets)
+        if split_dataset_from_train:
+            start, end = split_range
+            dataset_length = len(dataset)
+            start, end = round(start * dataset_length), round(end * dataset_length)
+            seed = new_config.split_train.seed
+            indices = np.random.RandomState(seed).permutation(dataset_length)[start:end]
```
apsdehal, line 149: Use torch to generate indices rather than np?

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,49 @@ def _download_requirement(
...
+        for imdb_idx in range(len(annotations)):
+            dataset_class = self.dataset_class
+            dataset = dataset_class(new_config, dataset_type, imdb_idx)
+            datasets.append(dataset)
+
+        dataset = MMFConcatDataset(datasets)
+        if split_dataset_from_train:
+            start, end = split_range
+            dataset_length = len(dataset)
+            start, end = round(start * dataset_length), round(end * dataset_length)
```
apsdehal, line 147: Make sure there is no overlap between two sets here.

apsdehal:
Formatting is failing on CircleCI, please fix that as well.

mmf/datasets/mmf_dataset_builder.py
ytsheng: This only calculates the start, end index for one dataset_type (ie., val, test, train).

mmf/datasets/mmf_dataset_builder.py
ytsheng: https://github.com/pytorch/pytorch/issues/17079 
It seems like the numpy interface is much cleaner compared to torch's.
Is there a reason that we prefer to use torch than numpy?

mmf/datasets/mmf_dataset_builder.py
apsdehal: Yes, i understand that, but they can overlap given the way they are calculated.

mmf/datasets/mmf_dataset_builder.py
apsdehal: Yeah, one extra thing for user to seed. We try to rely on torch as much as possible instead of numpy.

mmf/datasets/mmf_dataset_builder.py
ytsheng: added on the bottom

ytsheng:
Thanks for the review. I addressed your comments. I had to make some changes in the multi_dataset_loader due to the way empty dataset is handled (ie, if the val takes up 100% of the train, then the train will be of length 0).

mmf/datasets/multi_dataset_loader.py
```diff
@@ -132,17 +134,28 @@ def build_datasets(self, config):
...
+        if self._num_datasets:
+            self.current_index = 0
+            self.current_dataset = self.datasets[self.current_index]
 
         self._infer_dataset_probabilities()
+        self._did_build_datasets = True
 
     def build_dataloaders(self):
-        assert len(self._datasets) > 0, "Call build_datasets first"
+        if not self._did_build_datasets:
```
apsdehal, line 154: Why do we need this extra check? Maybe I am missing something here?

mmf/datasets/multi_dataset_loader.py
```diff
@@ -176,9 +189,14 @@ def _infer_dataset_probabilities(self):
             self._proportional_sampling = True
 
         if self._proportional_sampling is True and len(self._per_dataset_lengths) > 0:
-            self._dataset_probabilities = self._per_dataset_lengths[:]
+            dataset_lengths = self._per_dataset_lengths[:]
+            dataset_lengths = [
+                # Empty datasets need to have some probability to prevent infinite loop
+                max(1, dataset_len)
```
apsdehal, line 195: If they are already popped how would they reach this stage?

mmf/datasets/multi_dataset_loader.py
```diff
@@ -132,17 +134,28 @@ def build_datasets(self, config):
...
 
         self._infer_dataset_probabilities()
+        self._did_build_datasets = True
 
     def build_dataloaders(self):
-        assert len(self._datasets) > 0, "Call build_datasets first"
+        if not self._did_build_datasets:
+            assert len(self._datasets) > 0, "Call build_datasets first"
+
+        if len(self.datasets) == 0:
```
apsdehal, line 157: I think this logic is getting unnecessarily complicated for a single use case. We should fail fast when there is no datasets, or use try except python style.

apsdehal (changes requested):
Code readability has been impacted with the recent changes, it is hard to tell what is happening. See my comment on how to improve. Since, the feature for 0 size dataset is getting too complex for a simple feature, we should fail fast and prevent this. We shouldn't  avoid making these many changes in MultiDatasetLoader as it is most sensitive part of MMF.

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,61 @@ def _download_requirement(
...
+            if start > end:
+                raise ValueError(
+                    f"Train split ratio for {dataset_type} must be positive."
+                )
+
+            generator = torch.Generator()
+            generator.manual_seed(self.config.get("split_train.seed", 123456))
+            indices = torch.randperm(dataset_length, generator=generator)[start:end]
```
apsdehal, line 160: NIT: shift this to a separate function

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,61 @@ def _download_requirement(
...
+            dataset = MMFSubset(dataset, indices)
+        return dataset
+
+    def _modify_dataset_config_for_split(self, config):
+        for data_type in config.split_train:
+            if data_type == "seed":
+                continue
+            if config.use_images:
+                with open_dict(config.images):
```
apsdehal, line 169: NIT: You can open dict once on top.

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +123,61 @@ def _download_requirement(
...
+                    config.images[data_type] = copy.deepcopy(config.images.train)
```
apsdehal, line 170: NIT: Do `from copy import deepcopy`

mmf/datasets/multi_dataset_loader.py
```diff
@@ -132,17 +134,28 @@ def build_datasets(self, config):
...
             if hasattr(dataset_instance, "__len__"):
-                self._per_dataset_lengths.append(len(dataset_instance))
-                self._total_length += len(dataset_instance)
+                dataset_instance_length = len(dataset_instance)
+                if dataset_instance_length:
+                    self._per_dataset_lengths.append(dataset_instance_length)
+                    self._total_length += dataset_instance_length
+                else:
+                    # Filter out datasets with no items
+                    self.datasets.pop()
```
apsdehal, line 143: assert and fail here fast.

ytsheng:
Hi @apsdehal, thanks for the review! I addressed your comments and ran the tests. Here is the documentation: -

We explicitly assert when the dataset length is of size 0, this is demonstrated by the following dataset_config:
```
  textvqa:
    zoo_requirements:
    - textvqa.defaults
    - textvqa.ocr_en
    split_train:
      val: 0.9
      test: 0.1
    features:
      train:
      - textvqa/defaults/features/open_images/detectron.lmdb,textvqa/ocr_en/features/ocr_en_frcn_features.lmdb
   processors:
```
Here, we assert due to val + test ratio add up to 1, leaving 0 for training. However if you wish to convert most of train data to validation, you can do the following:
```
  textvqa:
    zoo_requirements:
    - textvqa.defaults
    - textvqa.ocr_en
    split_train:
      val: 0.99
      test: 0.001
    features:
      train:
      - textvqa/defaults/features/open_images/detectron.lmdb,textvqa/ocr_en/features/ocr_en_frcn_features.lmdb
      test:
      - test_path
   processors:
```
Here if you specified test's train split percentage, then we will use the specified percentage from train for test. However, if you don't specify test's train split percentage, we will attempt to use the data from the specified test_path. If there is no specified test_path and you don't specify the train_split percentage, the code will raise a ValueError.


I will add this new API change into the docs.

mmf/datasets/multi_dataset_loader.py
```diff
@@ -132,8 +132,10 @@ def build_datasets(self, config):
             self.datasets.append(dataset_instance)
 
             if hasattr(dataset_instance, "__len__"):
-                self._per_dataset_lengths.append(len(dataset_instance))
-                self._total_length += len(dataset_instance)
+                dataset_instance_length = len(dataset_instance)
+                assert dataset_instance_length
```
apsdehal, line 136: be more specific on why this assert failed with a message.

mmf/datasets/mmf_dataset_builder.py
```diff
@@ -119,6 +122,64 @@ def _download_requirement(
...
+                )
+            indices = self._generate_permuted_indexes(dataset_length)[start:end]
+            dataset = MMFSubset(dataset, indices)
+            print(
+                f"Dataset type: {dataset_type} length: {len(dataset)} total: {dataset_length}"
+            )
+        return dataset
+
+    def _generate_permuted_indexes(self, dataset_length):
+        import torch
```
apsdehal, line 164: Move to top.

apsdehal (approved):
This looks good to me to land. Please address minor comments before landing and update the API docs.

apsdehal:
Also, make sure to reimport the diff to phabrictor before shipping it off.

PR #477, opened 2020-08-13
[mmf] [feat] Batch Beam Search

howardhsu:
Summary:
Implement beam search for batch_size > 1.
Current Implementation of Beam Search works only for batch size = 1. Here is the implementation : https://github.com/facebookresearch/mmf/blob/08e5416db2ffa86508af4161505213b976df2cfd/mmf/utils/text.py#L271

Differential Revision: D22860178

### DIFF
--- a/mmf/utils/text.py
+++ b/mmf/utils/text.py
@@ -234,15 +234,18 @@ class TextDecoder:
         self._complete_seqs_scores = []
 
     def init_batch(self, sample_list):
+        img_size = sample_list.image_feature_0.size()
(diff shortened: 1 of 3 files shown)


vedanuj:
@howardhsu Can you please check the test failures as well as fix the linter failures?

vedanuj:
@howardhsu Can you fix the changes to the `test_model.py` file as that is causing the tests to fail for nucleus sampling. If you use case requires the change maybe add some condition?

PR #491, opened 2020-08-18
[fix] More informative logging for loading checkpoints.

suzyahyah:
Summary:
Previously unclear at what stage the loaded checkpoint file resumes training and this may have some unintended consequences.

Adding `num_updates`, `current_iteration`, and `current_epoch` to the logger.

Differential Revision: D23199211

### DIFF
--- a/mmf/utils/checkpoint.py
+++ b/mmf/utils/checkpoint.py
@@ -208,15 +208,18 @@ class Checkpoint:
                 self._load_optimizer(ckpt)
 
             self.trainer.early_stop_callback.early_stopping.init_from_checkpoint(ckpt)
-
-            logger.info("Checkpoint loaded")
-
             reset_counts = ckpt_config.reset.all or ckpt_config.reset.counts
 
             if not reset_counts:
                 self._load_counts(ckpt)
         else:
             self._load_pretrained(new_dict)
+
+        logger.info(
+            f"Checkpoint loaded.. num updates:{self.trainer.num_updates} "
+            + f"current iteration:{self.trainer.current_iteration} "
+            + f"current epoch:{self.trainer.current_epoch}"
+        )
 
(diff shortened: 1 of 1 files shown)


mmf/utils/checkpoint.py
```diff
@@ -208,16 +208,19 @@ def _load(self, file, force=False, load_zoo=False, load_pretrained=False):
...
-
             reset_counts = ckpt_config.reset.all or ckpt_config.reset.counts
 
             if not reset_counts:
                 self._load_counts(ckpt)
         else:
             self._load_pretrained(new_dict)
 
+        logger.info(
+            f"Checkpoint loaded.. num updates:{self.trainer.num_updates} "
```
apsdehal, line 219:
```suggestion
            f"Checkpoint loaded. Current num updates: {self.trainer.num_updates} "
```

mmf/utils/checkpoint.py
```diff
@@ -208,16 +208,19 @@ def _load(self, file, force=False, load_zoo=False, load_pretrained=False):
...
+            + f"current iteration:{self.trainer.current_iteration} "
```
apsdehal, line 220:
```suggestion
            + f"Current Iteration: {self.trainer.current_iteration} "
```

mmf/utils/checkpoint.py
```diff
@@ -208,16 +208,19 @@ def _load(self, file, force=False, load_zoo=False, load_pretrained=False):
...
+            + f"current epoch:{self.trainer.current_epoch}"
```
apsdehal, line 221:
```suggestion
            + f"Current epoch: {self.trainer.current_epoch}"
```

suzyahyah: Thanks for the pointers!

PR #506, opened 2020-08-21
[fix,feat] Add loss for in-batch cross entropy and allow "target" to be specified on the fly

suzyahyah:
[Converted to draft. Metrics still requires targets]

Summary:
Previously, losses require "targets" in sample_list, and will silently return `{}` if "targets" are not present in `sample_list`, but this might not be the case for custom loss functions.

Self-supervised/metric-learning/retrieval based functions like in batch cross entropy (example provided) or alignment losses compute their targets on the fly within the batch, and leaving this flexible allows for different targets to be specified for different loss functions on the fly.

Differential Revision: D23269219

### DIFF
--- a/mmf/modules/losses.py
+++ b/mmf/modules/losses.py
@@ -92,7 +92,6 @@ class Losses(nn.Module):
                     "sure that your ImDB has labels? you may have "
                     "wanted to run with evaluation.predict=true"
                 )
-            return output
 
         for loss in self.losses:
             output.update(loss(sample_list, model_output, *args, **kwargs))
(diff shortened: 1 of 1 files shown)


mmf/modules/losses.py
```diff
@@ -559,3 +558,20 @@ def __init__(self, params=None):
...
+        if params is None:
+            params = {}
+        self.loss_fn = nn.CrossEntropyLoss(**params)
+
+    def forward(self, sample_list, model_output):
+        bs = len(sample_list['id'])
+        scores = model_output['scores']
+        assert scores.shape[0] == scores.shape[1]
+        # our targets are just along the diagonal
+        targets = torch.arange(0, bs, dtype=torch.long).to(device=scores.device)
```
vedanuj, line 576: This seems to depend only on the batch size. I think this can be done in the model forward pass as well and you can add it to the sample_list.

PR #512, opened 2020-08-25
train retriever; model interface, retrieval training loop, and tuple dataset

suzyahyah:
Summary:
Extending MMBT for training a unimodal (I to I, T to T) and bimodal (I, T to I, T) retriever. I is Image and T is Text.

To train a retriever, we need to first construct a tuple dataset, which has anchor item and a reference to its positive item. The anchor item is loaded as per usual from dataloader, but the reference item can come from anywhere.

* New Dataset and Datasetbuilder, where we use `COCOTupleDataset` to demonstrate.

  * XTupleDataset, inherits from X Dataset but has an additional field `pos_image_ids` which needs to be prepared beforehand. These are its positive/similar image_ids (ref_ids).

  * A reference file which provides the mapping to the datapoint via the image_ids (ref_ids). The reference file does not need to be loaded by data loader and can come from anywhere.
`self.ref_data = self.construct_refdata()`

* Retriever Interface for any model in MMF to inherit, should implement `ref_encode_image` and `ref_encode_text`, which is used in unimodal and bimodal retrieval training.
  * `faim/mmf/mmf/models/interfaces/retriever.py`

  * `mmbt_rag` which inherits from `mmbt` and `RetrieverInterface` is used as an example here.

* New config files for `mm_rag`:
  * faim/mmf/projects/mm_rag/configs/mmbt
    * default_optimizer_retriever.yaml
    * rag_defaults.yaml
    * train_retriever.yaml

Limitations:

* While we have a way to train, we currently don't have a way to evaluate retrieval performance outside of batch training. Currently metrics like accuracy are tied to inbatch accuracy. Early stop is also tied to inbatch loss.
* For actual retrieval evaluation, we need to encode all items in the reference set into something like a faiss index and then calculate Precision Recall.
* The way the reference set is loaded in `faim/mmf/mmf/datasets/coco/dataset.py: COCOTupleDataset` is kind of ugly but we wanted a way to sample similar images randomly during training (not fix the anchor and positive item pair).

Differential Revision: D23304586

### DIFF
--- a/mmf/configs/datasets/vqa2/train_retriever.yaml
+++ b/mmf/configs/datasets/vqa2/train_retriever.yaml
@@ -1,0 +1,83 @@
+dataset_config:
+  coco_tuple:
+      data_dir: ${env.data_dir}/datasets
+      depth_first: false
+      fast_read: false
+      use_images: true
+      use_features: false
+      images:
+        train:
+        - coco/defaults/images/train2014
+        val:
+        - coco/defaults/images/val2014
+        test:
+        - coco/defaults/images/val2014
+      features:
+        train:
+        - coco/defaults/features/trainval2014.lmdb
+        val:
+        - coco/defaults/features/trainval2014.lmdb
+        test:
+        - coco/defaults/features/test2015.lmdb
+      annotations:
+        train:
+        - coco/defaults/annotations/imdb_karpathy_train_by_image_retrieve.jsonl
+        val:
+        - coco/defaults/annotations/imdb_karpathy_val_by_image_retrieve.jsonl
+        test:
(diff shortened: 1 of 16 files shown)


suzyahyah:
This is failing unit tests because it seems the new dataset key doesn't get registered

vedanuj:
It will be good to add more details about what kind of retrieval we are doing here. Also if you can add an example of how retrieval is done during training it will be useful to follow the code. By example, something like this : Text `T` and Image/Refs `I_1`, `I_2`, `I_3` etc. And the training task is to retrieve the `I_i` most relevant to the text `T`. 

You can add these details as docstrings to the new classes being added.

vedanuj:
Closing this as retrieval support has been added in #796

PR #513, opened 2020-08-25
[fix] load lr scheduler state dict during checkpoint loading

ronghanghu:
Summary: the current implementation does not load lr_scheduler's states from checkpoints, which is a bug (causing last_epoch to be reset when loading from checkpoints). This commit fixes it by saving and loading lr_scheduler's states. It fall backs to setting lr_scheduler's last_epoch when lr_scheduler's states don't exist in checkpoint.

Test Plan: tested with M4C checkpoint saving and loading.

(CPU tests failed due to black formatting error in mmf/datasets/processors/bert_processors.py, unrelated to this PR.)

### DIFF
--- a/mmf/trainers/mmf_trainer.py
+++ b/mmf/trainers/mmf_trainer.py
@@ -55,9 +55,12 @@ class MMFTrainer(
         self.lr_scheduler_callback = LRSchedulerCallback(self.config, self)
 
         # Add callbacks for execution during events
+        self.callbacks.append(self.lr_scheduler_callback)
+        # checkpoint_callback needs to be called after lr_scheduler_callback so that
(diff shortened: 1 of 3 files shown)


mmf/trainers/mmf_trainer.py
```diff
@@ -55,9 +55,12 @@ def configure_callbacks(self):
         self.lr_scheduler_callback = LRSchedulerCallback(self.config, self)
 
         # Add callbacks for execution during events
+        self.callbacks.append(self.lr_scheduler_callback)
```
apsdehal, line 58: Good catch.

ronghanghu:
@apsdehal I rebased onto master and imported it to Phabricator.

PR #514, opened 2020-08-26
[feat] Enable caption generation for all datasets registered to MMF

suzyahyah:
Summary:
Previously, caption generation could only be done for datasets which had “answers” as a data field and produced “scores”. This PR removes that dependency and should preserve all previous logic.

This PR allows the caption generation model to be applied to any dataset registered in MMF.

* removes the dependency on “answers” being in the text field in `utils/text.py` and `models/butd.py`.

* All datasets which inherit from mmf_dataset.py have a new function `format_for_caption_generation` 

* This function is modified from `mmf/datasets/builders/coco/dataset.py: format_for_caption_generation`, and preserves all keys with string `id`.

*`test_reporter.py` doesn’t require `scores` to be present for caption generation, and directly calls `format_for_caption_generation` if model is butd and in generation mode.

Differential Revision: D23339848

### DIFF
--- a/mmf/common/test_reporter.py
+++ b/mmf/common/test_reporter.py
@@ -151,7 +151,14 @@ class TestReporter(Dataset):
         if not is_master():
             return
 
-        results = self.current_dataset.format_for_prediction(report)
+        inference_type = None
+        if self.config.model == "butd":
+            inference_type = self.config.model_config.butd.inference.type
+
+        if inference_type in ["beam_search", "nucleus_sampling"]:
+            results = self.current_dataset.format_for_caption_generation(report)
+        else:
+            results = self.current_dataset.format_for_prediction(report)
 
         if hasattr(model, "format_for_prediction"):
             results = model.format_for_prediction(results, report)
--- a/mmf/configs/datasets/vqa2/for_butd_caption_generation.yaml
+++ b/mmf/configs/datasets/vqa2/for_butd_caption_generation.yaml
@@ -1,0 +1,36 @@
+dataset_config:
+  vqa2:
+    data_dir: ${env.data_dir}/datasets
+    depth_first: false
+    fast_read: false
+    use_images: false
+    use_features: true
+    images:
+      val:
+      - coco/defaults/images/train2014
+    features:
+      val:
+      - coco/defaults/features/trainval2014.lmdb
+    annotations:
+      val: 
+      - vqa2/defaults/annotations/imdb_train2014.npy
+    processors:
+      text_processor:
+        type: vocab
+        params:
+          max_length: 52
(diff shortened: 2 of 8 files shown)


mmf/configs/datasets/vqa2/for_butd_caption_generation.yaml
```diff
@@ -0,0 +1,36 @@
+dataset_config:
```
vedanuj, line 1: I don't think caption generation for VQA is required. VQA uses COCO images only.

mmf/datasets/mmf_dataset.py
```diff
@@ -50,6 +52,37 @@ def build_image_db(self):
         image_path = self._get_path_based_on_index(self.config, "images", self._index)
         return ImageDatabase(self.config, image_path, annotation_db=self.annotation_db)
 
+    def format_for_caption_generation(self, report):
```
vedanuj, line 55: Let's not add it here in MMFDataset. This should go to the individual datasets that require caption generation. For example for Hateful Memes, I would suggest add a separate dataset class for HM caption and add `format_for_prediction` in that which will generate caption predictions.

mmf/common/test_reporter.py
```diff
@@ -151,7 +151,14 @@ def add_to_report(self, report, model):
...
 
-        results = self.current_dataset.format_for_prediction(report)
+        inference_type = None
+        if self.config.model == "butd":
+            inference_type = self.config.model_config.butd.inference.type
+
+        if inference_type in ["beam_search", "nucleus_sampling"]:
+            results = self.current_dataset.format_for_caption_generation(report)
+        else:
+            results = self.current_dataset.format_for_prediction(report)
```
vedanuj, line 161: Let's keep only the `format_for_prediction` and remove this logic here. If report contains `captions` use that in `format_for_prediction` otherwise fall back to old implementation.

mmf/datasets/mmf_dataset.py
suzyahyah: So I was thinking of a use-case where any dataset registered to MMF can directly use the `butd` model for caption generation, or any other subsequent models for caption generation. So something like this would work immediately, as long as the model configs point to the butd model and the dataset configs point to that dataset.

`model=butd`
`dataset=<any-mmf-registered-dataset>`

mmf/common/test_reporter.py
suzyahyah: I agree, though I think a potential problem is that "caption" is an overloaded field, therefore it seems necessary to check for the inference type, and I am not sure if we can easily check that `model_config` in the dataset builders..

PR #518, opened 2020-08-27
[docs] FB internal documentation

vedanuj:
Differential Revision: D23365885

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -7,6 +7,13 @@
  * @format
  */
 
+function FBInternal(elements) {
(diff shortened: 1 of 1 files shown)


website/sidebars.js
```diff
@@ -7,6 +7,13 @@
  * @format
  */
 
+function FBInternal(elements) {
+  return process.env.FB_INTERNAL ? elements : [];
+}
+function FBInternalWithOssFallback(elements, fallback) {
```
apsdehal, line 13: Is this needed as of now, will throw a linting error on JS side if I run the linter.

PR #526, opened 2020-08-29
[fix] Fix lxmert model bbox processing

### LINKED ISSUE #520: problem in lxmert get_image_and_text_features fn
in get_image_and_text_features, position of box will be norm by following code 

https://github.com/facebookresearch/mmf/blob/3f313b6878d3d0cb55882415445ba31ef8bfa291/mmf/models/lxmert.py#L677-L687

in line 680, if image_h and image_w is provided, image_location will be init with zero? is there should be copy of bbox?

vedanuj:
- Fixes #520 and use `transformer_bbox_processor` processor
- `transformer_bbox_processor` will now work when bbox features already have area norm

Test Plan:

Run lxmert model for pretraining and finetuning on vqa2

### DIFF
--- a/mmf/models/lxmert.py
+++ b/mmf/models/lxmert.py
@@ -661,35 +661,15 @@ class LXMERT(BaseModel):
         image_info = getattr(sample_list, "image_info_0", {})
         image_dim_variable = getattr(image_info, "max_features", None)
         image_feature_variable = getattr(sample_list, "image_feature_0", None)
-        if image_feature_variable is None:
-            raise Exception("object features must be provided for lxmert")
-        bbox = getattr(image_info, "bbox", None)
-        if bbox is None:
-            raise Exception("bounding boxes must be provided for lxmert")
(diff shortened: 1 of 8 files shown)


mmf/models/lxmert.py
```diff
@@ -661,35 +661,15 @@ def get_image_and_text_features(self, sample_list, device):
...
-                image_location[:, :, 0] /= image_w[:, None]
-                image_location[:, :, 1] /= image_h[:, None]
-                image_location[:, :, 2] /= image_w[:, None]
-                image_location[:, :, 3] /= image_h[:, None]
-            else:
-                image_location = bbox
-            image_location_variable = image_location.to(device)
+        max_features = torch.tensor(
+            image_feature_variable.shape[1], dtype=torch.int
+        ).to(device)
```
apsdehal, line 666: device I believe can directly be specified using `device=device`

PR #529, opened 2020-09-02
[fix] Answer processor issues in MaskedVQA2Dataset

apsdehal:
- Adds add_answer option to default config
- Set it to default in the dataset
- Add default answer processor config

Test Plan:

Tested on full masked_vqa2 config under VisualBERT


vedanuj (approved):
Thanks for fixing this.

PR #545, opened 2020-09-04
[fix] Update HM docs for resume_pretrained=False during evaluation

apsdehal:
- Update README and docs


vedanuj (approved):
Thanks for updating this.

PR #550, opened 2020-09-08
 [feat] Add torchscript support for MMBT

vedanuj:
- Adds torchscript support for MMBT model
- Adds `BertEmbeddingsJit` and `BertModelJit` layers which are torchscriptable versions of `BertEmbeddings` and `BertModel` layers from transformers package
- Replaces transformers Bert model layers with scriptable layers by monkey patching, function `replace_with_jit`
- Adds tests

Test Plan:
- Run tests
- Run old pretrained checkpoints for MMBT and verify results

### DIFF
--- a/mmf/models/mmbt.py
+++ b/mmf/models/mmbt.py
@@ -7,17 +7,19 @@
 
 import os
 from copy import deepcopy
+from typing import Optional
 
 import torch
 from mmf.common.registry import registry
 from mmf.models.base_model import BaseModel
 from mmf.models.interfaces.mmbt import MMBTGridHMInterface
 from mmf.modules.encoders import MultiModalEncoderBase
+from mmf.modules.hf_layers import replace_with_jit
 from mmf.utils.checkpoint import load_pretrained_model
 from mmf.utils.configuration import get_mmf_cache_dir
 from mmf.utils.modeling import get_optimizer_parameters_for_bert
 from omegaconf import OmegaConf
-from torch import nn
+from torch import Tensor, nn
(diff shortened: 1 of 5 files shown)


mmf/models/mmbt.py
```diff
@@ -482,27 +473,42 @@ def __init__(self, config, *args, **kwargs):
...
         pooled_output = module_output[1]
         output = {}
 
-        if (
-            self.encoder_config.output_hidden_states
-            or self.encoder_config.output_attentions
-        ):
-            output["extras"] = module_output[2:]
+        if not torch.jit.is_scripting():
+            if (
```
apsdehal, line 503: Some comments from other PR are also applicable here, please take care of those.

apsdehal (commented):
Some of the comments from other PR are applicable here, please address those.

mmf/models/mmbt.py
```diff
@@ -227,90 +230,70 @@ def forward(
...
-            encoder_extended_attention_mask = encoder_attention_mask[:, None, None, :]
-
-        encoder_extended_attention_mask = encoder_extended_attention_mask.to(
-            dtype=next(self.parameters()).dtype
-        )  # fp16 compatibility
-        encoder_extended_attention_mask = (
-            1.0 - encoder_extended_attention_mask
-        ) * -10000.0
-
-        # Prepare head mask if needed
```
vedanuj, line 276: Removing this as we always set head_mask to None in the call to `MMBTModel`. Hence this is never used. Also scripting List[None] is not supported in the `BertEncoder`.

apsdehal (approved):
Probably need to check why CircleCI is having memory issues or disable that test for circleci cpu.

PR #551, opened 2020-09-08
[refactor] Refactor MMF Transformer embeddings to support scriptability

vedanuj:
- Changes to support torchscript for mmf transformer derivatives
- Changes to `MMFTransformerEmbeddings` by modifying the different modality embeddings to `nn.ModuleList` to make it scriptable.
- Some changes to the forward function 

Note: It is difficult to make MMF transformer scriptable as a whole since its base can be any type of transformer and making that scriptable is dependent on other libraries. The idea is to have derived models from MMF transformer that can be scripted. For example: We will have a version of MMF transformer that works with Bert, Roberta and XLMR that will be scriptable.

Test Plan:
- Test with hateful memes dataset

### DIFF
--- a/mmf/models/mmf_transformer.py
+++ b/mmf/models/mmf_transformer.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 from copy import deepcopy
-from typing import Any, Dict, Type
+from typing import Any, Dict, List, Type
 
 import torch
 from mmf.common.registry import registry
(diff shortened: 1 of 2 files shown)


mmf/models/mmf_transformer.py
```diff
@@ -287,31 +287,35 @@ def preprocess_sample(self, sample_list: Dict[str, Any]) -> BaseTransformerInput
 
         return BaseTransformerInput(input_ids, position_ids, segment_ids, masks)
 
-    def forward(self, sample_list: Dict[str, Any]) -> Dict[str, Tensor]:
+    def transformer_encoder(self, embedding, attention_mask):
```
apsdehal, line 290: The function names looks like a property. We should name it maybe `transfomer_encode` to avoid confusion.

mmf/models/mmf_transformer.py
```diff
@@ -182,6 +178,10 @@ def forward(
 class MMFTransformer(BaseTransformer):
     def __init__(self, config: BaseTransformerConfigType, *args, **kwargs):
         super().__init__(config)
+        self.classification_labels = self.config.num_labels
```
apsdehal, line 181: NIT: Name this `num_classfication_labels`.

mmf/models/transformers/base.py
```diff
@@ -19,6 +19,9 @@ class BaseTransformerInput:
     segment_ids: Dict[str, Tensor]  # dict of segment/token type ids for all modalities
     masks: Dict[str, Tensor]  # dict of masks for all modalities
 
+    def __getitem__(self, item):
```
apsdehal, line 22: Maybe add a comment, why this needs to be defined.

apsdehal (approved):
LGTM! Some NITs that you can fix before landing.

Issue #544, opened 2020-09-04
Cannont load 'bert-base-uncased' in transformer module: Weights from pretrained model not used in TextBert


xzChiang:
## ❓ Questions and Help
When I try to follow the experiment of M4C_Captioner Model:
```
CUDA_VISIBLE_DEVICES=0,1 mmf_run dataset=textcaps \
model=m4c_captioner \
config=projects/m4c_captioner/configs/m4c_captioner/textcaps/defaults.yaml \
env.save_dir=./save/m4c_captoiner \
run_type=train_val \
checkpoint.resume_zoo=m4c_captioner.textcaps.defaults \
env.data_dir=<data_dir_path>
```
Although it is finished, but the Log shows: 
```
| INFO | transformers.modeling_utils : loading weights file .../bert-base-uncased/pytorch_model.bin
[...]

PR #553, opened 2020-09-08
[fixup] fix config and zoo file for m4c_captioner

ronghanghu:
Summary:
- Fix the the mismatch between M4C-Captioner checkpoint and TextVQA features (#544). Keep two variants of M4C-Captioner config, one trained with newer features extracted with maskrcnn-benchmark (`defaults`), and the other trained with older features extracted with Caffe2 (`with_caffe2_feat`), which is used in our experimentations in the paper and has higher CIDEr.
- Create a MMF website page for M4C-Captioner.
- Also fix previous M4C website page.

Test plan:
- tested locally with model zoo checkpoints `m4c_captioner.textcaps.defaults` (for `defaults.yaml`) and `m4c_captioner.textcaps.with_caffe2_feat` (for `with_caffe2_feat.yaml`)


website/docs/projects/m4c_captioner.md
```diff
@@ -0,0 +1,110 @@
...
+
+Follow [checkpointing](https://mmf.sh/docs/tutorials/checkpointing) tutorial to understand more fine-grained details of checkpoint, loading and resuming in MMF
+
+:::
+
+Afterwards, use `projects/m4c_captioner/scripts/textcaps_eval.py` to evaluate the prediction json file. For example:
+```bash
+# the default data location of MMF (unless you have specified it otherwise)
+# this is where MMF datasets are stored
+export MMF_DATA_DIR=~/.cache/torch/mmf/data
```
apsdehal, line 103: I like the idea of specifying this explicitly.

website/docs/projects/m4c_captioner.md
```diff
@@ -0,0 +1,110 @@
+---
```
apsdehal, line 1: You need to add this to `sidebars.js`. Otherwise it won't load in the website.

apsdehal (approved):
LGTM! Thanks for fixing this. 
1) Please add the label to sidebars.js
2) Import to phabricator 
3) Check if the website loads properly in the jobs internally

PR #555, opened 2020-09-11
[refactor] feature extraction script upgrade

ytsheng:
* Update the comment to point to the correct maskrcnn-benchmark. There are many on github that is available. This one is the correct one to use.
* Added the X152 model
* Tested that both models can be loaded with the updated repo.

### DIFF
--- a/tools/scripts/features/extract_features_vmb.py
+++ b/tools/scripts/features/extract_features_vmb.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
-# Requires vqa-maskrcnn-benchmark to be built and installed
-# Category mapping for visual genome can be downloaded from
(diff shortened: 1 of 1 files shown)


vedanuj:
We should also to upload the latest models for X152 and update the comment, should be good to keep an option to use both the current(X101) as well as the X152 models. cc @apsdehal what do you think?

apsdehal:
Yes, that would be good addition as part of this PR.

vedanuj (commented):
Can we add an argument param that can help specify which model user wants to use? Also make sure both the models can be loaded with the updated repo (we only tested the X-152 model in that).

vedanuj:
Thanks for updating. Looks good! Can you import this to phabricator?

vedanuj (approved):

PR #558, opened 2020-09-12
[fix] Fix tests with larger CPU resource, fail-fast set to false

vedanuj:
- Increase resource size for CircleCI CPU tests, change to machine executor which have larger RAM
- set fail-fast to false for Actions fto be able to pinpoint which matrix combinations cause failures

Test Plan:
 - Run the tests and check if they pass


apsdehal (approved):
LGTM. Thanks.

PR #560, opened 2020-09-14
[fix] memory leak in loss reporting

suzyahyah:
Previously `meter_update_dict.update(reduced_loss_dict)` was storing the
loss with the gradient and computation graph. This PR should fix various GPU and CPU oom problems.

### DIFF
--- a/mmf/trainers/core/reporting.py
+++ b/mmf/trainers/core/reporting.py
@@ -30,14 +30,31 @@ class TrainerReportingMixin(ABC):
             # Add metrics to meter only when mode is `eval`
             meter_update_dict = {}
             if not eval_mode:
-                loss_key = report.dataset_type + "/total_loss"
-                reduced_loss = sum([loss.mean() for loss in reduced_loss_dict.values()])
-                if hasattr(reduced_loss, "item"):
-                    reduced_loss = reduced_loss.item()
+                total_loss_key = report.dataset_type + "/total_loss"
+                meter_update_dict, total_loss = self.update_dict(
+                    meter_update_dict, reduced_loss_dict
+                )
+                registry.register(total_loss_key, total_loss)
+                meter_update_dict.update({total_loss_key: total_loss})
 
-                registry.register(loss_key, reduced_loss)
-                meter_update_dict.update({loss_key: reduced_loss})
-                meter_update_dict.update(reduced_loss_dict)
             if hasattr(report, "metrics"):
-                meter_update_dict.update(reduced_metrics_dict)
+                meter_update_dict, _ = self.update_dict(
+                    meter_update_dict, reduced_metrics_dict
+                )
+
             meter.update(meter_update_dict, report.batch_size)
+
+    def update_dict(self, meter_update_dict, values_dict):
+        total_val = 0
+        for key, val in values_dict.items():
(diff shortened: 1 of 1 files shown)


suzyahyah:
not sure why this fails on circle CI, isort and black passed successfully during the commit using editorconfig..

vedanuj:
The circleci test failure is broken on master as well. Not related to this change

vedanuj (approved):
Thanks for fixing this. Looks good to me!

mmf/trainers/core/reporting.py
```diff
@@ -37,7 +37,10 @@ def update_meter(
 
                 registry.register(loss_key, reduced_loss)
                 meter_update_dict.update({loss_key: reduced_loss})
-                meter_update_dict.update(reduced_loss_dict)
+                for key in reduced_loss_dict:
+                    loss = reduced_loss_dict[key].item()
+                    meter_update_dict[key] = loss
```
apsdehal, line 42: Do it in single line?

mmf/trainers/core/reporting.py
```diff
@@ -37,7 +37,10 @@ def update_meter(
...
+
             if hasattr(report, "metrics"):
                 meter_update_dict.update(reduced_metrics_dict)
```
apsdehal, line 45: We should this for metrics as well if metric hasattr item to be foolproof? In rare cases, this may also blow out when metric actually doesn't do .item itself.

apsdehal (changes requested):

mmf/trainers/core/reporting.py
suzyahyah: > We should this for metrics as well if metric hasattr item to be foolproof? In rare cases, this may also blow out when metric actually doesn't do .item itself.

Ok, in the latest version I refactored updating `reduced_metrics_dict` and `reduced_loss_dict` to a single update function which checks for `.item()`.

mmf/trainers/core/reporting.py
```diff
@@ -30,14 +30,30 @@ def update_meter(
...
             if hasattr(report, "metrics"):
-                meter_update_dict.update(reduced_metrics_dict)
+                meter_update_dict, _ = self.update_dict(
+                    meter_update_dict, reduced_metrics_dict
+                )
+
             meter.update(meter_update_dict, report.batch_size)
+
+    def update_dict(self, meter_update_dict, values_dict):
+
```
apsdehal, line 48: is this new line needed?

mmf/trainers/core/reporting.py
```diff
@@ -30,14 +30,30 @@ def update_meter(
...
+        total_val = 0
+        for key, val in values_dict.items():
+            if val.dim() == 1:
```
apsdehal, line 51: You should first actually check if `dim` attribute is present in val. As for meters, this can be a scalar.

apsdehal (commented):
Thank you for making the changes, there are some things that will still fail.

mmf/trainers/core/reporting.py
suzyahyah: got it, we now check if `torch.is_tensor` before calling .dim()

PR #561, opened 2020-09-14
[feat] Upload model weights to LXMERT

dowhilefalse:
Hi! thanks so much for the help and adding lxmert to MMF project, I would like to upload the pretrained models weights that can be used for downstream training tasks.

here is the google drive link that I store models in:
https://drive.google.com/drive/folders/1ev-ZGjx0e6Mf5e6fbhHnoc1R5Nf46rNf?usp=sharing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,18 +1,9 @@
 # Use 2.1 for orbs
 version: 2.1
 
-# Include windows orb for windows bui
-orbs:
-  win: circleci/windows@2.2.0
-
 # -------------------------------------------------------------------------------------
 # Environments to run the jobs in
 # -------------------------------------------------------------------------------------
-cpu: &cpu
-  machine:
-    image: ubuntu-1604:201903-01
-  resource_class: large
-
 gpu: &gpu
   environment:
     CUDA_VERSION: "10.2"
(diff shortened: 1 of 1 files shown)


vedanuj:
Hi @dowhilefalse .. Thanks for the model. Can you give more details about which dataset this model is pretrained on? Also was it pretrained using the lxmert implementation in mmf? Would also be good if yuo can provide finetuned weights for downstream tasks like VQA or GQA.

dowhilefalse:
Hi, thanks for the feedback! 
this model is pretrained on joint datasets of coco, vqa2, gqa, and visual_genome, using the original lxmert implementation.

Yes sure, I will upload the finetuned weights very soon.

dowhilefalse:
Hi! I just uploaded vqa and gqa finetuned weights (vqa_downstream, gqa_downstream) in the google drive.

and also note that the pretrained weights (gqa_bert_state_dict, vqa_bert_state_dict, nlvr_bert_state_dict) for three downstream tasks are different.

Thanks so much! Let me know if you need anything else.

NohTow:
Hello, 
I was trying to use pre-trained LXMERT with mmf, yet the original weights are not compatible with mmf so your weights will be a better fit.
Can you elaborate on this part : "_and also note that the pretrained weights (gqa_bert_state_dict, vqa_bert_state_dict, nlvr_bert_state_dict) for three downstream tasks are different._"
The original paper does not mention different version of pre-training as far as I know, shouldn't be the pre-training task agnostic ? Which weights correspond to the "original pre-training" ? Is it related to the downstream task heads ?

Thanks in advance and sorry if it is not the place to discuss that, but I think these precisions can benefit everyone that will use these weights in the future.

vedanuj:
Hi @dowhilefalse .. we tried to load these models in mmf but some keys couldn't be loaded. Can you make sure the models can be loaded? Additionally you can upload the models on your server and update the [model.yaml](https://github.com/facebookresearch/mmf/blob/master/mmf/configs/zoo/models.yaml) zoo file so the checkpoints can be used in mmf. 

Please update the PR with changes to `models.yaml` as well.

dowhilefalse:
Hi @NohTow thanks! if you refer to the original implementation of LXMERT, https://github.com/airsplay/lxmert,  you can see that we load classifier weights for VQA and GQA but not for NLVR. And also there is some process to convert the weights dictionary.

PR #562, opened 2020-09-16
[feat] Added the localized narratives dataset

ytsheng:
1. Added the masked language modeling task for localized narratives using just caption data.
2. Still in the process of uploading the features file to s3 storage so that we can reference it here.


ytsheng:
Creating draft PR for now

PR #564, opened 2020-09-16
[refactor] Use config object for ImageFeatureEncoder init

apsdehal:
Summary: There has been a mismatch between ImageFeatureEncoder and ImageEncoder signatures which doesn't allow us to create same config types for them even though that's how they are used. This is confusing and should be same. This diff aims to fix this discrepancy.

Differential Revision: D23712213

### DIFF
--- a/mmf/models/m4c.py
+++ b/mmf/models/m4c.py
@@ -7,8 +7,9 @@ import torch
 import torch.nn.functional as F
 from mmf.common.registry import registry
 from mmf.models.base_model import BaseModel
-from mmf.modules.encoders import ImageFeatureEncoder
 from mmf.modules.layers import ClassifierLayer
+from mmf.utils.build import build_image_encoder
+from omegaconf import OmegaConf
 from torch import nn
 from transformers.modeling_bert import (
     BertConfig,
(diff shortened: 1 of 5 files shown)


mmf/models/m4c.py
```diff
@@ -44,6 +45,19 @@ def build(self):
...
 
+    def _build_encoder_config(self):
+        return OmegaConf.create(
+            {
+                "type": "finetune_faster_rcnn_fpn_fc7",
+                "params": {
+                    "in_dim": 2048,
+                    "weights_file": "models/detectron.defaults/fc7_w.pkl",
+                    "bias_file": "models/detectron.defaults/fc7_b.pkl",
+                    "model_data_dir": self.config.model_data_dir,
```
vedanuj, line 56: While you are here, maybe we can move these(`in_dim`, `weights_file`, `bias_file`) values to config?

mmf/models/movie_mcan.py
```diff
@@ -68,12 +68,10 @@ def _init_feature_encoders(self, attr: str):
         feature_dim = self.config[attr + "_feature_dim"]
         setattr(self, attr + "_feature_dim", feature_dim)
 
-        encoder_type = feat_encoder.type
-        encoder_kwargs = copy.deepcopy(feat_encoder.params)
-        encoder_kwargs.model_data_dir = self.config.model_data_dir
-        encoder_kwargs.cond_features = self.text_embeddings_out_dim
```
ytsheng, line 74: Do we intend to remove the `cond_features`?
If we remove it here, can we also remove the cond_features in the movie_mscan's config yaml file?

mmf/models/m4c.py
apsdehal: I didn't wanted to add this so as not to break older configs. That's why I skipped it for m4c.

vedanuj: If there aren't a lot of m4c configs maybe it's better to move these there now, otherwise might cause problems later on if someone wants to change these values.

vedanuj (approved):
Looks good!

mmf/models/m4c.py
apsdehal: Ok, let me handle this separately and create a task for it.

PR #565, opened 2020-09-16
[feat] Add init dataclasses for mmbt and encoders

apsdehal:
Summary:
As step one of FAIM integration, we allow building our models from config so that the models are purely decoupled from configuration and users know what the model expects.

We first do this for the MMBT model.

- Adds configs for MMBT and respective encoders.
- Also adds from_params method for MMBT

There is an issue with OmegaConf that doesn't let us use Union in structured configs. Take a look at https://github.com/omry/omegaconf/issues/144

Differential Revision: D23699688

### DIFF
--- a/mmf/models/m4c.py
+++ b/mmf/models/m4c.py
@@ -7,8 +7,9 @@ import torch
 import torch.nn.functional as F
 from mmf.common.registry import registry
 from mmf.models.base_model import BaseModel
-from mmf.modules.encoders import ImageFeatureEncoder
 from mmf.modules.layers import ClassifierLayer
+from mmf.utils.build import build_image_encoder
+from omegaconf import OmegaConf
 from torch import nn
 from transformers.modeling_bert import (
     BertConfig,
@@ -44,6 +45,19 @@ class M4C(BaseModel):
         self._build_mmt()
         self._build_output()
 
+    def _build_encoder_config(self):
+        return OmegaConf.create(
+            {
+                "type": "finetune_faster_rcnn_fpn_fc7",
+                "params": {
+                    "in_dim": 2048,
+                    "weights_file": "models/detectron.defaults/fc7_w.pkl",
+                    "bias_file": "models/detectron.defaults/fc7_b.pkl",
+                    "model_data_dir": self.config.model_data_dir,
+                },
+            }
+        )
+
     def _build_txt_encoding(self):
         TEXT_BERT_HIDDEN_SIZE = 768
 
@@ -77,12 +91,8 @@ class M4C(BaseModel):
 
     def _build_obj_encoding(self):
         # object appearance feature: Faster R-CNN
-        self.obj_faster_rcnn_fc7 = ImageFeatureEncoder(
-            encoder_type="finetune_faster_rcnn_fpn_fc7",
-            in_dim=2048,
-            weights_file="models/detectron.defaults/fc7_w.pkl",
-            bias_file="models/detectron.defaults/fc7_b.pkl",
-            model_data_dir=self.config.model_data_dir,
+        self.obj_faster_rcnn_fc7 = build_image_encoder(
+            self._build_encoder_config(), direct_features=True
         )
         # apply smaller lr to pretrained Faster R-CNN fc7
         self.finetune_modules.append(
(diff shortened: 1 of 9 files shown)


mmf/modules/encoders.py
```diff
@@ -12,14 +15,44 @@
...
+
+# SuperClass for encoder params
+@dataclass
+class EncoderParams:
+    pass
+
+
+@dataclass
+class EncoderConfig:
+    type: str = MISSING
```
vedanuj, line 39: Just curious, what does `MISSING` do compared to leaving it without any value? And if a value is required why not use `MISSING` everywhere in all the dataclass additions in this PR?

mmf/modules/encoders.py
```diff
@@ -103,9 +148,27 @@ def forward(self, image):
...
+@dataclass
+class ResNet152ImageEncoderParams(EncoderParams):
+    pretrained: bool = True
+    # "avg" or "adaptive"
+    pool_type: str = "avg"
+    num_output_features: int = 1
+
+
+@dataclass
+class ResNet152ImageEncoderConfig(ImageEncoderConfig):
```
vedanuj, line 160: Do you think it is better to make the 'Config' classes as nested inside the main class? For example `ResNet152ImageEncoderConfig` can be a nested class inside `ResNet152ImageEncoder`.

vedanuj (commented):
Thanks for adding these. Have a few comments.

mmf/models/m4c.py
```diff
@@ -44,6 +45,19 @@ def build(self):
         self._build_mmt()
         self._build_output()
 
+    def _build_encoder_config(self):
```
ytsheng, line 48: May I suggest that we move this default config class `_build_encoder_config` to be within the mmf.utils.build class? It is the same with the one in this PR: #564. Perhaps we could add in arguments to the original `build_image_encoder` so that it can be configurable. If the arguments don't exist, we can use those defaults. :)

apsdehal: Hey, 
1) This is change is present here because this was exported from stacked diff. #564 is the first diff in that stack and this change is part of it.
2) This is usually not the default image encoder everywhere. So, I would refrain from adding this as default for now. Instead, identity would suit better as default image encoder. It would be better to fail with an error that config was not passed, instead of initializing with something that user doesn't expect by default, IMO.

mmf/modules/encoders.py
apsdehal: MISSING means that value will need to be passed by the user in all cases. I will review again to see if I am missing "MISSING" anywhere it should be present.

mmf/modules/encoders.py
apsdehal: This is a very good point for discussion. I was thinking about this based on what PyText does. Then, on the other hand, this is style in which transformer does it. Let me think a little more about this. I do agree that keeping Config objects inside the actual module using it is better as it keep it grouped, but on the other side it also increases some friction in reusability.

vedanuj:
Looks good overall. There are some changes to MMBT model as well. Please update the description to reflect those.

mmf/models/mmbt.py
```diff
@@ -7,24 +7,34 @@
...
 from mmf.utils.modeling import get_optimizer_parameters_for_bert
-from omegaconf import OmegaConf
+from omegaconf import II, DictConfig, OmegaConf
 from torch import Tensor, nn
 from transformers.modeling_bert import BertForPreTraining, BertPredictionHeadTransform
 
 
 # TODO: Remove after transformers package upgrade to 2.5
-class MMBTConfig:
+class MMBTConfigForTransformers:
```
vedanuj, line 37: We should keep this as `MMBTConfig`. These were copied from Huggingface and once we bump up the internal version we can replace these without changes to other parts of the code.

mmf/models/mmbt.py
```diff
@@ -504,9 +528,41 @@ def forward(self, sample_list: Dict[str, Tensor]):
...
+        use_modal_end_token: bool = True
+        fused_feature_only: bool = False
+        output_dim: int = 768
+
+    def __init__(self, config: Union[DictConfig, Config], *args, **kwargs):
         super().__init__(config)
 
+    @classmethod
+    def from_params(cls, **kwargs):
+        return MMBT(OmegaConf.structured(cls.Config(**kwargs)))
```
vedanuj, line 564: So for this we don't require the base model config yaml file anymore is that right? Should we remove it to avoid confusion? (I am worried as there are two sources of truth for default configurations for the MMBT model now)

vedanuj (approved):

mmf/models/mmbt.py
apsdehal: This is something I was discussing with Omry and he has a solution for. Anyway, removing this can end up as a breaking change. We can remove this altogether once we move to hydra.

mmf/models/mmbt.py
apsdehal: Makes sense.

PR #568, opened 2020-09-17
[feat] Add torchscriptable version for MMF Transformer

vedanuj:
Summary:

- Modifications to MMF transformer `preprocess_sample` to be scriptable friendly.
- Refactor out backends : Added a `BaseTransformerBackend` abstract class.
- Added a new scriptable backend for MMF Transformer that supports only Huggingface Bert, Roberta and XLMR transformer models.  Here we are implementing a version of MMF Transformer that supports torchscriptable layers and works only for BERT, RoBERTa and XLMR models
- Add scriptable `RobertaEmbeddingsJIT`
- Add tests for MMF Transformer

Differential Revision: D23569352

### DIFF
--- a/mmf/common/registry.py
+++ b/mmf/common/registry.py
@@ -21,6 +21,7 @@ Various decorators for registry different kind of classes with unique keys
 - Register a optimizer: ``@registry.register_optimizer``
 - Register a scheduler: ``@registry.register_scheduler``
 - Register a decoder: ``@registry.register_decoder``
+- Register a transformer backend: ``@registry.register_transformer_backend``
 """
 from mmf.utils.env import setup_imports
 
@@ -47,6 +48,7 @@ class Registry:
         "scheduler_name_mapping": {},
         "processor_name_mapping": {},
         "decoder_name_mapping": {},
+        "transformer_backend_name_mapping": {},
         "state": {},
     }
 
@@ -266,6 +268,14 @@ class Registry:
         return wrap
 
     @classmethod
+    def register_transformer_backend(cls, name):
+        def wrap(func):
+            cls.mapping["transformer_backend_name_mapping"][name] = func
+            return func
+
+        return wrap
+
+    @classmethod
     def register_decoder(cls, name):
         r"""Register a decoder to registry with key 'name'
 
@@ -353,6 +363,10 @@ class Registry:
     @classmethod
     def get_decoder_class(cls, name):
         return cls.mapping["decoder_name_mapping"].get(name, None)
+
+    @classmethod
+    def get_transformer_backend_class(cls, name):
+        return cls.mapping["transformer_backend_name_mapping"].get(name, None)
 
     @classmethod
     def get(cls, name, default=None, no_warning=False):
--- a/mmf/configs/models/mmf_transformer/defaults.yaml
+++ b/mmf/configs/models/mmf_transformer/defaults.yaml
@@ -2,11 +2,17 @@ model_config:
   mmf_transformer:
     transformer_base: bert-base-uncased
     training_head_type: classification
+    backend:
+      type: huggingface
+      params: {}
     num_labels: 2
     modalities:
       - type: text
         key: text
+        position_dim: 512
         segment_id: 0
+        layer_norm_eps: 1e-12
+        hidden_dropout_prob: 0.1
       - type: image
         key: image
         embedding_dim: 2048
--- a/mmf/models/base_model.py
+++ b/mmf/models/base_model.py
@@ -170,8 +170,10 @@ class BaseModel(nn.Module):
             assert isinstance(
                 model_output["losses"], collections.abc.Mapping
             ), "'losses' must be a dict."
-        else:
+        elif hasattr(self, "losses"):
             model_output["losses"] = self.losses(sample_list, model_output)
+        else:
+            model_output["losses"] = {}
 
         return model_output
 
--- a/mmf/models/mmf_transformer.py
+++ b/mmf/models/mmf_transformer.py
@@ -1,7 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
-from copy import deepcopy
-from typing import Any, Dict, List, Type
+from typing import Dict, List
 
 import torch
 from mmf.common.registry import registry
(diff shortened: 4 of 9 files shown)


mmf/common/registry.py
```diff
@@ -21,6 +21,7 @@
 - Register a optimizer: ``@registry.register_optimizer``
 - Register a scheduler: ``@registry.register_scheduler``
 - Register a decoder: ``@registry.register_decoder``
+- Register a transformer backend: ``@registry.register_transformer_backend``
```
apsdehal, line 24: Suggestion: I think it would be better to not use registry for this as it might need to be used independent of registry. I would suggest that we follow the route of ImageEncoder style factory pattern and provide a `build_transformer_backend` function which can then return the proper backend module. This is something that will probably stay in core for quite some time and thus should be easier handled via that API as it won't require calling `setup_imports` and other things required to initialize MMF. We should think of this as one of the module that MMF provides.

mmf/configs/models/mmf_transformer/defaults.yaml
```diff
@@ -2,11 +2,16 @@ model_config:
   mmf_transformer:
     transformer_base: bert-base-uncased
     training_head_type: classification
+    backend:
+      type: huggingface
```
apsdehal, line 6: Maybe we should also keep a params property here.

mmf/models/base_model.py
```diff
@@ -170,7 +170,7 @@ def __call__(self, sample_list, *args, **kwargs):
             assert isinstance(
                 model_output["losses"], collections.abc.Mapping
             ), "'losses' must be a dict."
-        else:
+        elif hasattr(self, "losses"):
```
apsdehal, line 173: Does this cause issues? This might cause issues down the line in the trainer where it won't find the losses key to sum upon. One option is to add a final case where you add empty dict for value key "losses" here?

mmf/models/mmf_transformer.py
```diff
@@ -193,133 +57,111 @@ def build_encoders(self):
...
-        )
-
     def build_heads(self):
         """Initialize the classifier head. It takes the output of the
         transformer encoder and passes it through a pooler (we use the pooler from BERT
         model), then dropout, BertPredictionHeadTransform (which is a linear layer,
         followed by activation and layer norm) and lastly a linear layer projecting the
         hidden output to classification labels.
         """
+        transformer_config = self.backend.get_config()
```
apsdehal, line 67: So, here's a thought, I don't know specifics of pytext but I think something like this would be better:

- Initialize the backend in init
- backend has `get_pooler` function similar to get_config() function, then you don't need to pass config to it specifically and it can stay backend agnostic
- Similarly, config values can become properties of backend class which will be agnostic again to backend.
- Same for prediction head transform.
- This can be extended to other BERT module layers as well. 
- Now backend also handle cases which are not regular like difference in xlm-r and bert etc in the hidden shadows itself. 

If something is missing in a model, backend can then return identity class.

mmf/models/transformers/backends/huggingface.py
```diff
@@ -0,0 +1,225 @@
...
+                    segment_ids[modality_name]
+                )
+
+            list_embeddings.append(dropout(layer_norm(total_embedding)))
+
+        return torch.cat(list_embeddings, dim=1)
+
+
+@registry.register_transformer_backend("huggingface")
+class HuggingfaceBackend(nn.Module):
```
apsdehal, line 159: I think there should be base backend class which defines specific properties and functions which are required by MMFTransformer and other models so that the code can stay agnostic to the model itself.

apsdehal: Then, you would use the backend to build multiple part of the model. This is also fine, but this should also help initialize other layers.

mmf/modules/hf_layers.py
```diff
@@ -31,6 +30,7 @@ def replace_with_jit():
     BertSelfAttention.transpose_for_scores = BertSelfAttentionJit.transpose_for_scores
     BertModel.forward = BertModelJit.forward
     PreTrainedModel.__ignored_properties__ = ["base_model", "dummy_inputs"]
+    RobertaEmbeddings.forward = RobertaEmbeddingsJit.forward
```
apsdehal, line 33: As we have seen this monkey patching inside helpful internally, we need to find a way to patch embeddings through backend maybe.

apsdehal (changes requested):

mmf/models/mmf_transformer.py
vedanuj: I think for these layers in the `head` it's fine to get the config and build it in the main model itself. These are not BERT specific(although they are taken from bert model). Additionally all the different backends don't provide any specific layers for classifiers or poolers. So for those who don't provide we will require calling a different backend(which I think is cumbersome). They make these changes in the model forward pass itself. So in order to keep the forward return type same for all backends, the pooler or classifier layers should  be built in the main model IMO.

mmf/modules/hf_layers.py
vedanuj: The `hf_layers` contains all the Huggingface layers only and `replace_with_jit` is called from that backend. Once we add the MMF Transformer to Fluent2 we can make the change as required.

mmf/common/registry.py
vedanuj: I made the change to ImageEncoder style. However, if having these in registry is not an issue it's better to have them in registry. Let me know your thoughts and I revert that change.

mmf/models/transformers/backends/huggingface.py
```diff
@@ -0,0 +1,218 @@
...
+                total_embedding += self.token_type_embeddings(
+                    segment_ids[modality_name]
+                )
+
+            list_embeddings.append(dropout(layer_norm(total_embedding)))
+
+        return torch.cat(list_embeddings, dim=1)
+
+
+class HuggingfaceBackend(BaseTransformerBackend):
```
apsdehal, line 159: Here now that they inherit from same class, you can add them to registry. This is similar to how each model inherits from BaseModel.

This is needed because internal files would require try except without registry pattern.

mmf/models/transformers/backends/huggingface.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def forward(
+        self,
+        tokens_ids: Dict[str, Tensor],
+        position_ids: Dict[str, Tensor],
+        segment_ids: Dict[str, Tensor],
+        masks: List[Tensor],
+    ) -> Tuple[Tensor, List[Tensor]]:
+        # Transformer Input Embeddings
+        embedding_output = self.embeddings(
```
apsdehal, line 203: This should be moved under `generate_embeddings` function where each backend do more stuff on their embeddings output (transpose etc.) if needed.

mmf/models/transformers/backends/huggingface.py
```diff
@@ -0,0 +1,218 @@
...
+            input_ids=tokens_ids, position_ids=position_ids, segment_ids=segment_ids
+        )
+
+        # Transformer Attention mask
+        attention_mask = torch.cat(masks, dim=-1)
+        extended_attention_mask = attention_mask.unsqueeze(1).unsqueeze(2)
+        extended_attention_mask = (1.0 - extended_attention_mask) * -10000.0
```
apsdehal, line 210: Similarly, this can be moved under `generate_attention_mask`

mmf/models/transformers/backends/huggingface.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Transformer Encoder
+        encoded_layers = self._transformer_encode(
+            embedding_output,  # combined embedding
+            extended_attention_mask,  # combined attention mask
+        )
```
apsdehal, line 216: `generate_encoded_layers(embedding_output, attention_mask)` You can pass more stuff in signature to be more generic. But you also can always modified the signature in child class.

All these functions can then go into BaseBackend class

mmf/models/transformers/backends/huggingface.py
```diff
@@ -0,0 +1,218 @@
...
+            list_embeddings.append(dropout(layer_norm(total_embedding)))
+
+        return torch.cat(list_embeddings, dim=1)
+
+
+class HuggingfaceBackend(BaseTransformerBackend):
+    """Transformer backend wih Huggingface transformer models
+    """
+
+    def __init__(self, config: BaseTransformerConfigType, *args, **kwargs):
```
apsdehal, line 163: Please add docstrings as well to the functions.

mmf/models/transformers/base.py
```diff
@@ -53,24 +45,62 @@ class BaseTransformerConfigType:
...
+        Warning: Empty shell for code to be implemented in other class.
+        """
+
+    @abstractmethod
+    def get_config(self):
+        """Return the transformer configuration. This can be the config built
+        in `build_transformer_config` or the model config passed to init.
+
+        Warning: Empty shell for code to be implemented in other class.
+        """
```
apsdehal, line 84: raise NotImplementedError in these.

vedanuj: Since these are abstract methods, will complain anyway if not implemented.

apsdehal (approved):
LGTM! This is a great step forward for MMF.

PR #572, opened 2020-09-18
[fix] Black formatting in sample

apsdehal:
- Accidentally forgot to run formatter on sample.py
- This fixes it


vedanuj (approved):
I think the PR description is wrong. Please fix it before landing.

PR #575, opened 2020-09-21
[WIP][feat] Add pretraining MLM head to mmf transformer

vedanuj:
MMF Transformer pretraining head


vedanuj:
Will open a new one.

PR #580, opened 2020-09-25
[feat] check unused model parameters in optimizer

ronghanghu:
This PR adds optimizer option on whether to allow some of the model's parameters not to be used by the optimizer. Default is false to guard against missing parameters due to implementation errors in a model's `get_optimizer_parameters`. Previously `get_optimizer_parameters` is error-prong as the users need to remember to update `get_optimizer_parameters` whenever adding a new submodule, and the error is hard to spot when the users forget to do so.

Test plan: tested locally w/ VisualBERT, against correctly and
incorrectly implemented `get_optimizer_parameters`.

### DIFF
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -161,7 +161,11 @@ run_type: train_inference
 
 # Configuration for optimizer, examples can be found in models' configs in
 # configs folder
-optimizer: {}
+optimizer:
+    # Whether to allow some of the model's parameters not to be used by the
+    # optimizer. Default is false to guard against missing parameters due to
+    # implementation errors in a model's get_optimizer_parameters
+    allow_unused_parameters: false
 
 # Configuration for scheduler, examples can be found in models' configs
 scheduler: {}
--- a/mmf/utils/general.py
+++ b/mmf/utils/general.py
@@ -143,7 +143,33 @@ def get_optimizer_parameters(model, config):
     if is_parallel and hasattr(model.module, "get_optimizer_parameters"):
         parameters = model.module.get_optimizer_parameters(config)
 
+    for group in parameters:
+        group["params"] = list(group["params"])
+
+    check_unused_parameters(parameters, model, config)
+
     return parameters
+
+
+def check_unused_parameters(parameters, model, config):
+    optimizer_param_set = {p for group in parameters for p in group["params"]}
+    unused_param_names = []
+    for n, p in model.named_parameters():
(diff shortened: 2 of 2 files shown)


vedanuj:
Thanks for this. We should also fix the visual_bert model if any params are not added to optimizer. Did you see any improvement in accuracy after adding those params to optimizer?

ronghanghu:
> Thanks for this. We should also fix the visual_bert model if any params are not added to optimizer. Did you see any improvement in accuracy after adding those params to optimizer?

The MMF VisualBERT's `get_optimizer_parameters` implementation is correct! I just tried manually introduce some bugs into it to make sure it gets detected by this PR 😃

mmf/utils/general.py
```diff
@@ -143,9 +143,36 @@ def get_optimizer_parameters(model, config):
...
+    check_unused_parameters(parameters, model, config)
+
     return parameters
 
 
+def check_unused_parameters(parameters, model, config):
+    optimizer_param_set = {p for group in parameters for p in group["params"]}
+    unused_param_names = [
+        n
+        for n, p in model.named_parameters()
```
apsdehal, line 158: Would it be possible to write this in simple for loop format? It kinda hard to read as of now.

mmf/configs/defaults.yaml
```diff
@@ -161,7 +161,11 @@ run_type: train_inference
 
 # Configuration for optimizer, examples can be found in models' configs in
 # configs folder
-optimizer: {}
+optimizer:
+    # Whether to allow some of the model's parameters not to be used by the
+    # optimizer. Default is false to guard against missing parameters due to
+    # implementation errors in a model's get_optimizer_parameters
+    allow_unused_parameters: false
```
apsdehal, line 168: Thoughts on whether this can break any of the existing models?

apsdehal (approved):
LGTM! Thanks for fixing this long standing issue. Left two minor comments. Have a look and import this to Phabricator.

mmf/utils/general.py
ronghanghu: I rewrote it in a for-loop (and tested it again).

mmf/configs/defaults.yaml
ronghanghu: I don't think it breaks our current models, but it may affect probably some custom-implemented user models that implement getting parameters other than get_optimizer_parameters. For example, a generative adversarial network where discriminator parameters are fetched separately and fed into a separate optimizer.

However, users can always set `optimizer.allow_unused_parameters` to True to bypass this check.

PR #605, opened 2020-10-05
[fixup] don't init early_stopping from ckpt when resetting counts

ronghanghu:
Since the docs in mmf/configs/defaults.yaml on `reset.count` mention that "All counts such as best_update, current_iteration etc will be reset", the early stopping criteria (`best_iteration` and `best_metric_value`) should not be initialized from the ckpt when `reset.count` is on.

Test plan: tested internally on a DETR-related model when initializing from an earlier model's checkpoint.


apsdehal (approved):
LGTM! Please import.

vedanuj (approved):

PR #609, opened 2020-10-05
[feat] First e2e test with emphasis on user dir

apsdehal:
This PR introduces first e2e test which uses `mmf_cli.run` programmatic
API and log parsing to check for log file. Specifically in this PR we
test for user_dir import by creating our own simple model and a dataset
that always generate one.

For a more friendly `user_dir` support, now we override default config
in configuration with user_cmd_opts as well so that MMF can discover
configs relative to user dir if needed.

We run the usual run command and expect the model to overfit and always
return 1.

Test Plan:

Tested locally on devfair using `pytest -v ./tests`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ eggs/
 .vscode/*
 *.so
 *-checkpoint.ipynb
+!tests/data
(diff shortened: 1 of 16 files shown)


vedanuj (approved):
Thanks for addressing the comments. Looks good to me.

PR #616, opened 2020-10-08
[feat] Introduce QC tests for missing init and empty folders 

apsdehal:
- Missing inits cause problem with user dir and packaging 
- The tests check for missing init in folders with python files
- Also, checks for empty folders
- Furthers, checks on all MMF folders can be added easily

Test Plan:
Tested locally and fixed the issues found

TODO:

Need to test internally as well.


vedanuj (approved):
LGTM!

PR #618, opened 2020-10-09
[chores] Remove cpu and windows build from CircleCI tests

vedanuj:
- Remove CPU tests and Windows Build test from CircleCI as they are already covered in Actions
  - Some CircleCI CPU tests were flaky
  - These tests also require high resources(class `large`) that causes forked repo builds to fail

Test Plan:
Check if tests run properly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,18 +1,9 @@
 # Use 2.1 for orbs
 version: 2.1
 
-# Include windows orb for windows bui
-orbs:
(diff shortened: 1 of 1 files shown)


apsdehal (approved):
LGTM! In long term, if GPU solution arrives, we should just get rid of CircleCI.

apsdehal (commented):
One benefit of having this was that circleci allowed us to ssh into windows machine and debug the cause of some failures there. This is something that I would miss :'(

PR #620, opened 2020-10-10
[fix] Fix optimizer params to include all modules for transformer models

vedanuj:
Recent changes in PR #568 moved the `pooler` layer outside classifier and was not added to optimizer params group.  Check added in #580 helped to catch this. Thanks @ronghanghu ! 

- PR refactors the code in `get_optimizer_parameters_for_bert`
- **All** layers other than classifier will have LR of `(lr * finetune_lr_multiplier)` when model head is not of `pretraining` type.
- Remove unnecessary setting LR when `finetune_lr_multiplier == 1`

Test Plan:
- Checked with MMF Transformer, MMBT, VisualBERT

### DIFF
--- a/mmf/utils/modeling.py
+++ b/mmf/utils/modeling.py
@@ -1,7 +1,11 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
+import logging
+
 from torch import nn
 
+
+logger = logging.getLogger(__name__)
 
 ACT2FN = {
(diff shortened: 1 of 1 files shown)


apsdehal (approved):
We should add tests for this in a separate PR to make sure this doesn't happen again, initializing VB, MMFT, MMBT in a test and making sure that get_optimizer_parameters returns all parameters. Approving this to land and unblock for now

PR #622, opened 2020-10-12
[fix] Change format of concatenating image and text as bert input in MMBT

BruceChaun:
Summary:
THe previous version of concatenation of image and text tokens has format `<cls> image_token <sep> <cls> text_tokens <sep> <pad>`, but there is a bug in code -- the `<sep>` token behind `image_token` might be `<pad>` because it simply used the last token in `input_ids`. We can easily see `bert_processor` convert input text into tokens which has format `<cls> tokens <sep> <pad>`.

Apart from resolving the aforementioned issue, this diff also converts to a simpler format `<cls> image_token <sep> text_tokens <sep> <pad>`. So based on previous version, modifications are
1. Extract the last non-masked token, which is `<sep>` as `modal_end_token`
2. Remove `<cls>` token at the beginning of text input ids and append 0 at the end to ensure it has the same sequence length
3. Fix `input_mask` by truncating the first token and appending 0 at the end

Differential Revision: D24164033

### DIFF
--- a/mmf/models/mmbt.py
+++ b/mmf/models/mmbt.py
@@ -335,6 +335,33 @@ class MMBTBase(MultiModalEncoderBase):
 
         self.mmbt = MMBTModel(self._mmbt_config, text_encoder, modal_encoder)
 
+    def extract_modal_end_token(self, sample_list: Dict[str, Tensor]):
+        # compute the position of the last non-masked token, which is <sep>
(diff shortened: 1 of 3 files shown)


apsdehal (approved):
I checked this diff on existing MMBT checkpoints, here are the results:

- with images: ROC_AUC: 67.67
- with features: ROC_AUC: 62.78

We need to retrain and upload both models again to zoo. 

But we can land this PR for now. We can update the zoo later.

PR #624, opened 2020-10-13
[feat] Enable torchscript on full VisualBERT model

apsdehal:
- This removes opinions on module getting torchscripted rather than the
full model. This will also enable us to keep the signature of Dict[str,
Tensor] same for all of the MMF models
- Modifies code flow according to scripting model and pretraining model
conditions
- Add helps function around getattr
- Modifies the test to play nice with new API which looks better

Test Plan:
Modifies VisualBERT tests for TorchScript to script and test the whole
model

### DIFF
--- a/mmf/models/visual_bert.py
+++ b/mmf/models/visual_bert.py
@@ -14,6 +14,7 @@ from mmf.modules.embeddings import BertVisioLinguisticEmbeddings
 from mmf.modules.hf_layers import BertEncoderJit, BertLayerJit
 from mmf.utils.configuration import get_mmf_cache_dir
 from mmf.utils.modeling import get_optimizer_parameters_for_bert
+from mmf.utils.torchscript import getattr_torchscriptable
 from mmf.utils.transform import (
     transform_to_batch_sequence,
     transform_to_batch_sequence_dim,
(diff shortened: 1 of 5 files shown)


mmf/models/visual_bert.py
```diff
@@ -416,123 +412,134 @@ def build(self):
...
-                )
+    def add_post_flatten_params(
+        self, sample_list: Dict[str, Tensor]
+    ) -> Dict[str, Tensor]:
+        sample_list["visual_embeddings_type"] = torch.zeros_like(
+            sample_list["image_mask"]
+        )
 
-        if sample_list.image_mask is not None:
+        if sample_list["image_mask"] is not None:
```
vedanuj, line 435: Since the previous `if sample_list.image_mask is not None:` is removed can we also remove this `if-else`?

mmf/models/visual_bert.py
```diff
@@ -545,26 +552,31 @@ def format_state_key(cls, key):
...
         )
 
-        if "pretraining" in self.config.training_head_type:
+        if self.training_head_type == "pretraining" and not torch.jit.is_scripting():
             loss_key = "{}/{}".format(
-                sample_list.dataset_name, sample_list.dataset_type
+                sample_list["dataset_name"], sample_list["dataset_type"]
             )
             output_dict["losses"] = {}
             output_dict["losses"][loss_key + "/masked_lm_loss"] = output_dict.pop(
```
vedanuj, line 582: I think we should throw an error if `torch.jit.is_scripting() and self.training_head_type == "pretraining"`. Reason is we are not updating the loss in the output dict properly here along with dataset name and type, which might cause issues during multi task pretraining. So basically we won't support scripting mode during pretraining.

vedanuj (commented):
Looks good. Left a couple of comments.

mmf/models/visual_bert.py
apsdehal: Makes sense. I actually thought that there is nothing stopping you from actually doing pretraining in scripted mode except dataset_name and dataset_type so I kept the support for this open. This will require modifications to tests as well. I will make this change and update the diff.

vedanuj (approved):
LGTM!

PR #629, opened 2020-10-15
[feat] Upload model weights to LXMERT

dowhilefalse:
Hi! Sorry I accidentally closed the PR.

@vedanuj , yes I tried to load weights to my server and add it to model.yaml. But it was not successful as the model did not seem to auto-download or locate the path.

But I tried to find out the problem you might be asking about so I made some changes to the model code. And updated the weights in google drive. https://drive.google.com/drive/folders/1ev-ZGjx0e6Mf5e6fbhHnoc1R5Nf46rNf?usp=sharing

I also updated the model.yaml. if this looks correct. 

Thanks so much!

### DIFF
--- a/mmf/configs/zoo/models.yaml
+++ b/mmf/configs/zoo/models.yaml
@@ -1,3 +1,40 @@
+lxmert:
+  pretrained:
+    vqa2:
+      defaults:
+        version: 1.0
+        resources:
+        - url: 
+          file_name: vqa_pretrained.pth
+    gqa:
+      defaults:
+        version: 1.0
+        resources:
+        - url: 
+          file_name: gqa_pretrained.pth
+          
+    nlvr:
+      defaults:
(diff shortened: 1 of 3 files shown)


mmf/configs/zoo/models.yaml
```diff
@@ -1,3 +1,40 @@
+lxmert:
+  pretrained:
+    vqa2:
+      defaults:
+        version: 1.0
+        resources:
+        - url:
```
vedanuj, line 7: You will have to add the url of the file in your server from where the models can be downloaded. You can check the other examples in this file.

dowhilefalse: I added the url in my test. I also tried other models like vilbert, but still didn't automatically download their weights.
Is there any thing to specify in the config yaml files?

PR #630, opened 2020-10-15
[fix] Parameter list regression in build_optimizers

apsdehal:
- With recent change of #580, there is a regression for models which
don't have `get_optimizer_parameters` or don't return param groups. #580
specifically expects param groups to be present.
- This PR fixes this regression and add tests for optimizers so that
this doesn't happen again

Test Plan:

Tested added for two type of parameter groups returned from the model

### DIFF
--- a/tests/modules/test_optimizers.py
+++ b/tests/modules/test_optimizers.py
@@ -1,0 +1,34 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import unittest
+
+import torch
+from mmf.models.mmbt import MMBT
+from mmf.utils.build import build_optimizer
+from omegaconf import OmegaConf
+from tests.test_utils import SimpleModel, skip_if_no_network
+
+
+class TestOptimizers(unittest.TestCase):
+    def setUp(self):
+        self.config = OmegaConf.create(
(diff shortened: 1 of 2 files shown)


ronghanghu (approved):
This looks good to me. Thanks for fixing it!

tests/modules/test_optimizers.py
```diff
@@ -0,0 +1,34 @@
...
+    def test_build_optimizer_simple_model(self):
+        model = SimpleModel(1)
+
+        optimizer = build_optimizer(model, self.config)
+        self.assertTrue(isinstance(optimizer, torch.optim.Optimizer))
+        self.assertEqual(len(optimizer.param_groups), 1)
+
+    @skip_if_no_network
+    def test_build_optimizer_custom_model(self):
+        model = MMBT.from_params()
```
vedanuj, line 27: Since the PR is fixing for models that don't contain `get_optimizer_parameters` we should also probably test one of those.

vedanuj (approved):

tests/modules/test_optimizers.py
apsdehal: Yes, that's what SimpleModel test is for.

PR #631, opened 2020-10-15
[feat] Make top level loss classes torchscriptable

apsdehal:
- We weren't doing init_losses in the tests so we didn't notice this
- Changes make top level loss function torchscriptable so that if losses
are empty, this doesn't cause any issues

Test Plan:

Enable init_losses in mmbt tests

### DIFF
--- a/tests/models/test_mmbt.py
+++ b/tests/models/test_mmbt.py
@@ -3,7 +3,6 @@
 import unittest
 
 import tests.test_utils as test_utils
-from mmf.common.registry import registry
 from mmf.models.mmbt import MMBT
 from mmf.modules.encoders import (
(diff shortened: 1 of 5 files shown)


vedanuj (approved):

tests/models/test_mmbt.py
```diff
@@ -33,6 +33,7 @@ def setUp(self):
         config.model_config[model_name]["num_labels"] = 2
         self.finetune_model = model_class(config.model_config[model_name])
         self.finetune_model.build()
+        self.finetune_model.init_losses()
```
vedanuj, line 36: Can we add this to MMF transformer tests as well?

PR #634, opened 2020-10-16
[fix] Issues with encoder update, add tests

apsdehal:
- I introduced some issues with recent encoder update. This PR aims to
fix that.
- This PR also adds tests to make sure this doesn't happen in future.
- Tests only tests initialization as of now, can be build upon in
future.

Test Plan:
Tests have been added

### DIFF
--- a/tests/modules/test_encoders.py
+++ b/tests/modules/test_encoders.py
@@ -1,0 +1,48 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import tempfile
+import unittest
+
+from mmf.modules import encoders
+from tests.test_utils import setup_proxy
+from torch import nn
+
+
+class TestEncoders(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/modules/test_encoders.py
```diff
@@ -0,0 +1,49 @@
...
+import tempfile
+import unittest
+
+from mmf.modules import encoders
+from tests.test_utils import skip_if_no_network
+from torch import nn
+
+
+class TestEncoders(unittest.TestCase):
+    def _test_init(self, cls, **params):
```
vedanuj, line 12: We can use proxy to enable running these tests internally as well. You can call  `test_utils.setup_proxy()` here (or better once in a `setUp` method). Then we can remove the `@skip_if_no_network`.

vedanuj (approved):
LGTM!

PR #636, opened 2020-10-19
[feat] Add optimizer state sharding (ZeRO)

vedanuj:
Summary:
Adding Zero optimizer state sharding to MMF from [fairscale](https://github.com/facebookresearch/fairscale) library.
Added to tutorials

It can used with this option : `optimizer.enable_state_sharding=True`

Differential Revision: D24317858

### DIFF
--- a/mmf/utils/build.py
+++ b/mmf/utils/build.py
@@ -226,7 +226,24 @@ def build_optimizer(model, config):
             )
 
     parameters = get_optimizer_parameters(model, config)
-    optimizer = optimizer_class(parameters, **params)
+
+    if optimizer_config.get("enable_state_sharding", False):
+        # TODO(vedanuj): Remove once OSS is moved to PT upstream
+        try:
+            from fairscale.optim.oss import OSS
+        except ImportError:
+            print(
+                "Optimizer state sharding requires fairscale. "
+                + "Install using pip install fairscale."
+            )
+            raise
+
+        assert (
+            is_dist_initialized()
+        ), "Optimizer state sharding can only be used in distributed mode."
+        optimizer = OSS(params=parameters, optim=optimizer_class, **params)
+    else:
+        optimizer = optimizer_class(parameters, **params)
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -22,6 +22,7 @@ install_dep: &install_dep
       command: |
         source activate mmf
         pip install --upgrade setuptools
+        pip install --progress-bar off -r requirements.txt
```
apsdehal, line 25: I am somewhat not a fan of this as it creates confusion. `setup.py` should in itself be enough to install any dependencies.

mmf/utils/build.py
```diff
@@ -226,7 +226,16 @@ def build_optimizer(model, config):
             )
 
     parameters = get_optimizer_parameters(model, config)
-    optimizer = optimizer_class(parameters, **params)
+
+    if optimizer_config.get("enable_state_sharding", False):
+        from fairscale.optim.oss import OSS
```
apsdehal, line 231: We can also do a `try`, `except ImportError` and ask user to install the fairscale library separately if we don't want explicit dependency on it. This would also solve some things on fbcode side, user who want it should include fairscale in the targets. Adding more dependency always causes more pain.

apsdehal (commented):
This is awesome. Some comments. Do you plan to add any tests for these?

mmf/utils/build.py
```diff
@@ -226,7 +226,24 @@ def build_optimizer(model, config):
             )
 
     parameters = get_optimizer_parameters(model, config)
-    optimizer = optimizer_class(parameters, **params)
+
+    if optimizer_config.get("enable_state_sharding", False):
```
apsdehal, line 230: It would make sense to write a simple test to test this. We have had issue with `build_optimizer` in past.

apsdehal (approved):
This looks good. Thanks for pushing this. A minor comment about test and should be good to land after that.

PR #638, opened 2020-10-19
[fix] Fix in_dim for movie_mcan model

vedanuj:
`in_dim` needs to be added to the config after the change in PR #564. Fixing that.


apsdehal (approved):

apsdehal (commented):
If possible, we should add simple init test for MoViE+MCAN

apsdehal (approved):

PR #642, opened 2020-10-19
[enhancement] Add freeze backend option to MMFT

vedanuj:
Summary: Add option to freeze backend of MMF Transformer model

Differential Revision: D24381850

### DIFF
--- a/mmf/models/transformers/base.py
+++ b/mmf/models/transformers/base.py
@@ -42,7 +42,6 @@ class BaseTransformerConfigType:
     token_noise_mean: float  # stddev of normal noise for token embeddings
     layer_norm_weight_fill: float  # layer norm weight initialization
     random_initialize: bool  # random initialize whole network
-    freeze_transformer: bool  # freeze the base transformer
     finetune_lr_multiplier: float  # finetune lr multiplier for base transformer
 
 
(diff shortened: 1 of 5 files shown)


projects/mmf_transformer/configs/vqa2/defaults.yaml
```diff
@@ -45,4 +45,6 @@ training:
 
 checkpoint:
   pretrained_state_mapping:
-    transformer: transformer
+    pooler: pooler
+    backend.transformer: backend.transformer
```
apsdehal, line 49: Won't `backend: backend` work?

mmf/models/transformers/base.py
```diff
@@ -178,6 +177,10 @@ def build_backend(self):
         backend_class = registry.get_transformer_backend_class(backend_type)
         self.backend = backend_class(self.config)
 
+        if getattr(backend_config, "freeze", False):
```
apsdehal, line 180: Can you instead do `backend_config.get("freeze", False)`

apsdehal (approved):
Thanks for this. Added two comment. Resolve before landing.

projects/mmf_transformer/configs/vqa2/defaults.yaml
vedanuj: It will work. However I want to keep them separate as we don't always want to initialize the embeddings from the pretrained models (due to different number of modalities between pretrained and finetuned tasks).

PR #644, opened 2020-10-19
[fix] Allow struct=True configs for MMFT encoders,pretrained Pythia

apsdehal:
- If modality is missing encoder key, MMFT after recent fix won't break
down
- This allows using configs without specifying encoder key and if they
are in struct mode which is what MMF gives when ran from command line

Test Plan:

Tested with audio video MMFT

### DIFF
--- a/mmf/models/pythia.py
+++ b/mmf/models/pythia.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import copy
 
+import omegaconf
 import torch
 from mmf.common.registry import registry
 from mmf.models.base_model import BaseModel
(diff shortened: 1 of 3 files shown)


mmf/models/pythia.py
```diff
@@ -81,8 +82,9 @@ def _init_feature_encoders(self, attr):
 
         for feat_encoder in feat_encoders_list_config:
             feat_encoder_config = copy.deepcopy(feat_encoder)
-            feat_encoder_config.params.model_data_dir = self.config.model_data_dir
-            feat_encoder_config.params.in_dim = feature_dim
+            with omegaconf.open_dict(feat_encoder_config):
+                feat_encoder_config.params.model_data_dir = self.config.model_data_dir
+                feat_encoder_config.params.in_dim = feature_dim
```
vedanuj, line 87: We should make this change for `movie_mcan` model as well.

vedanuj (approved):

mmf/models/pythia.py
apsdehal: Done. Thanks for pointing out.

PR #655, opened 2020-10-21
[fix] Fix text encoder output if the output is tensor

BruceChaun:
Summary: Some text encoder outputs return a single tensor (say with dimension `batch_size x out_dim`). Additional condition checks are added to make sure this case is not affected.

Differential Revision: D24456401

### DIFF
--- a/mmf/models/fusions.py
+++ b/mmf/models/fusions.py
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+import collections
 from copy import deepcopy
 
 import torch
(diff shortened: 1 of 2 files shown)


vedanuj (approved):

mmf/models/fusions.py
```diff
@@ -44,7 +44,7 @@ def forward(
 
         # Case of bert encoder, we only need pooled output. For BertModelJIT encoder
         # pooled output is the 2nd in the tuple(sequence, pooled, encoded_layers)
-        if len(text) >= 2:
+        if not torch.is_tensor(text) and isinstance(text, tuple) and len(text) >= 2:
```
apsdehal, line 47: Thinking more we should compare with `isinstance(text, collections.abc.Sequence)` instead of just `isinstance(text, tuple)`.

apsdehal (changes requested):

mmf/models/fusions.py
apsdehal: This would cover cases where the text is a list and other iterables. I confirmed this is not true for torch.Tensor.

apsdehal (approved):
Thanks for the change. Please do check and confirm in your runs that this works.

PR #659, opened 2020-10-23
[fix] use torch current device rather than using registry

ytsheng:
* Minor fix so that we remove the registry current device dependency, and instead use torch.

### DIFF
--- a/mmf/datasets/base_dataset.py
+++ b/mmf/datasets/base_dataset.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 from mmf.common.registry import registry
 from mmf.common.sample import SampleList
+from mmf.utils.general import get_current_device
 from torch.utils.data.dataset import Dataset
 
 
@@ -24,7 +25,7 @@ class BaseDataset(Dataset):
         self._dataset_name = dataset_name
         self._dataset_type = dataset_type
         self._global_config = registry.get("config")
-        self._device = registry.get("current_device")
+        self._device = get_current_device()
         self.use_cuda = "cuda" in str(self._device)
 
     def load_item(self, idx):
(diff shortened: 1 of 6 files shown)


mmf/datasets/base_dataset.py
```diff
@@ -24,7 +25,11 @@ def __init__(self, dataset_name, config, dataset_type="train", *args, **kwargs):
         self._dataset_name = dataset_name
         self._dataset_type = dataset_type
         self._global_config = registry.get("config")
-        self._device = registry.get("current_device")
+        self._device = (
+            f"cuda:{torch.cuda.current_device()}"
+            if torch.cuda.is_available()
+            else torch.device("cpu")
+        )
```
vedanuj, line 32: This can be made into an util function, since I see this being used in multiple places.

mmf/datasets/base_dataset.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 from mmf.common.registry import registry
 from mmf.common.sample import SampleList
+from mmf.utils.general import get_device
```
apsdehal, line 4: Probably name it `get_current_device`, `get_device` can refer to any device?

mmf/utils/general.py
```diff
@@ -316,3 +316,11 @@ def assert_iterator_finished(iter):
         pass
     else:
         assert False
+
+
+def get_device():
+    return (
```
apsdehal, line 322: Since this is a function now, can we write it is as a simple if else condition now?

mmf/utils/general.py
```diff
@@ -316,3 +316,11 @@ def assert_iterator_finished(iter):
         pass
     else:
         assert False
+
+
+def get_device():
```
apsdehal, line 321: Even though this is small, it would still be good to write tests for this to catch any pytorch related regressions in future.

apsdehal (changes requested):
Thanks. This is looking good. Some minor comments and should be good to go after that.

apsdehal (approved):

Issue #652, opened 2020-10-21
Training of BUTD produced very bad results


netw0rkf10w:
Hello,

I tried training the BUTD model using the command provided in the [documentation](https://mmf.sh/docs/projects/butd/):

```
mmf_run config=projects/butd/configs/coco/defaults.yaml \
    model=butd \
    dataset=coco \
    run_type=train
```

I didn't change anything except two minor configuration values in `defaults.yaml`:

```
training:
  tensorboard: true
  num_workers: 10
```

The results are very very bad at the end, the best validation BLEU4 is 0.021671, which means the model didn't seem to learn anything:

`2020-10-21T07:49:06 | mmf.trainers.callbacks.logistics: progress: 50000/50000, val/coco/caption_cross_entropy: 2.0662, val/total_loss: 2.0662, val/coco/caption_bleu4: 0.0187, num_updates: 50000, epoch: 31, iterations: 50000, max_updates: 50000, val_time: 54s 586ms, best_update: 1000, best_iteration: 1000, best_val/coco/caption_bleu4: 0.021671`

TensorBoard output:

<img width="1030" alt="Screen Shot 2020-10-21 at 16 43 13" src="https://user-images.githubusercontent.com/8569515/96736094-8b7df800-13bc-11eb-9d8c-5b620ee44756.png">

Could you please help me on this?

Thank you very much in advance!

netw0rkf10w:
I've tried running the evaluation using the provided pretrained model from the model zoo, and the results are also very bad: `val/coco/caption_bleu4: 0.0263`. It seems there's a bug in the computation of the metrics.

I also observed this warning:

`WARNING 2020-10-21T17:55:07 | py.warnings: /gpfsdswork/projects/rech/upf/ugt86fu/code/mmf/mmf/common/report.py:82: UserWarning: captions not found in report. Metrics calculation might not work as expected.`

netw0rkf10w:
@vedanuj Could you please have a look at this issue that seems to be critical?

netw0rkf10w:
Something is very wrong!

Running the provided pre-trained model on the validation set, I observed that the output json contains captions that start with either "a person that is standing..." or "a person sitting on..." for all the images!

vedanuj:
What is the Bleu score you are getting after running the pretrained model on the COCO validation set?

netw0rkf10w:
@vedanuj I obtained `0.0263`, as mentioned in my previous comment. It seems something is wrong in the decoding...

[...]

PR #664, opened 2020-10-28
[feat,fix] Stronger checkpoint validation, fix missing keys

### LINKED ISSUE #661: Performance difference while running M4C pretrained model
## ❓ Questions and Help

I ran the M4C pretrained model following the M4C documentation using the following command:

`mmf_run dataset=textvqa   model=m4c   config=projects/m4c/configs/textvqa/joint_with_stvqa.yaml   env.save_dir=./save/m4c   run_type=val   checkpoint.resume_zoo=m4c.textvqa.with_stvqa training.num_workers=1
`

I am getting a validation accuracy of 0.3882 as compared to 0.4055 which is reported in the paper. Is this difference in performance expected?

apsdehal:
- This PR aims to add stronger checkpoint validations to MMF
- It will fail the run if there are missing keys in the checkpoint
- It will log unexpected keys as warnings
- Using this, we fix multiple missing keys in pythia, m4c

Fixes #661 #652

### DIFF
--- a/mmf/utils/checkpoint.py
+++ b/mmf/utils/checkpoint.py
@@ -194,7 +194,27 @@ class Checkpoint:
         state_dict = self.upgrade_state_dict(ckpt["model"])
 
         if len(pretrained_state_mapping.items()) == 0:
-            self.trainer.model.load_state_dict(state_dict, strict=False)
+            incompatible_keys = self.trainer.model.load_state_dict(
+                state_dict, strict=False
+            )
+            if len(incompatible_keys.missing_keys) != 0:
+                logger.warning(
+                    f"Missing keys {incompatible_keys.missing_keys} in the"
+                    + " checkpoint.\n"
+                    + "If this is not your checkpoint, please open up an "
+                    + "issue on MMF GitHub. \n"
+                    + f"Unexpected keys if any: {incompatible_keys.unexpected_keys}"
+                )
+
+            if len(incompatible_keys.unexpected_keys) != 0:
+                logger.warning(
+                    "Unexpected keys in state dict: "
+                    + f"{incompatible_keys.unexpected_keys} \n"
+                    + "This is usually not a problem with pretrained models, but "
+                    + "if this is your own model, please double check. \n"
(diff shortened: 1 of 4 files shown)


mmf/utils/checkpoint.py
```diff
@@ -194,7 +194,25 @@ def _load(self, file, force=False, load_zoo=False, load_pretrained=False):
         state_dict = self.upgrade_state_dict(ckpt["model"])
 
         if len(pretrained_state_mapping.items()) == 0:
-            self.trainer.model.load_state_dict(state_dict, strict=False)
+            incompatible_keys = self.trainer.model.load_state_dict(
+                state_dict, strict=False
+            )
+            assert len(incompatible_keys.missing_keys) == 0, (
```
vedanuj, line 200: Can we show a warning for this instead of assert? Sometimes some extra params get added to models. Recently I discovered with latest version of transformers that they have added a new variable to store position ids. That will make our old checkpoints throw error.

mmf/utils/checkpoint.py
```diff
@@ -320,18 +338,26 @@ def upgrade_state_dict(self, state_dict):
...
-                new_dict[new_attr.replace("module.", "", 1)] = state_dict[attr]
+                new_attr = new_attr.replace("module.", "", 1)
             elif data_parallel and not attr.startswith("module."):
-                new_dict["module." + new_attr] = state_dict[attr]
-            else:
-                new_dict[new_attr] = state_dict[attr]
+                new_attr = "module." + new_attr
+
+            # Log if key has changed but not when it is only
+            # data parallel difference
```
vedanuj, line 357:
```suggestion
            # Log if key has changed but not when the difference
            # is due to data parallel `module.`
```

vedanuj (approved):
Looks good. Leaving a couple of comments.

mmf/utils/checkpoint.py
apsdehal: Hmm, I do think this is good assert validation, but yeah, warning sounds good.

PR #666, opened 2020-10-29
[chores] Upgrade transformers dependency for mmf

vedanuj:
Summary:
- Includes changes for compatibility with transformers 3.4.0
- BertLayerNorm is replaced with LayerNorm from torch.nn
- Roberta classes are replicated in the latest version. So we need to monkey patch all of them.
- BertEmbedding has a new position_id variable, due to which existing checkpoints have to be loaded with strict=False. This doesn't effect the models, instead of generating position ids it is just an added variable. In most of our models its a no-op.
- Additional properties need to be ignored in the `PreTrainedModel` class.

Differential Revision: D24599639

### DIFF
--- a/mmf/models/base_model.py
+++ b/mmf/models/base_model.py
@@ -41,6 +41,7 @@ Example::
 
 
 import collections
+import logging
 import warnings
 from copy import deepcopy
 from dataclasses import dataclass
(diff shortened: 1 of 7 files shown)


ytsheng (approved):
This looks good with me. I think we should add the test for Roberta jit, what do you think?

mmf/models/base_model.py
```diff
@@ -224,7 +224,7 @@ def from_pretrained(cls, model_name, *args, **kwargs):
         instance = cls(config)
         instance.is_pretrained = True
         instance.build()
-        instance.load_state_dict(checkpoint)
+        instance.load_state_dict(checkpoint, strict=False)
```
apsdehal, line 227: We should probably print like we did in other checkpoint case on what was unexpected and what was missing.

apsdehal (commented):
Thanks for this. Did we confirm if there are any regressions on current model with new version that can prevent us from upgrading?

PR #675, opened 2020-11-03
[fix,feat] update the state_key based on module

ytsheng:
- Allow for module-wise state dict key update
- Make use of the `_register_load_state_dict_pre_hook` to update the key of the state dict
  - opted for this approach because recursion is already being implemented in the load_state_dict function and therefore I think there is no need to re-implement recursion. Better to make use of the pytorch implementation.
- Slightly cleaner fix compared to this fix: [664](https://github.com/facebookresearch/mmf/pull/664)
- Some documentation clean up

### DIFF
--- a/mmf/modules/encoders.py
+++ b/mmf/modules/encoders.py
@@ -125,6 +125,32 @@ class FinetuneFasterRcnnFpnFc7(ImageFeatureEncoder):
         self.lc.bias.data.copy_(torch.from_numpy(bias))
         self.out_dim = out_dim
 
+    def _load_from_state_dict(
+        self,
+        state_dict,
+        prefix,
+        local_metadata,
+        strict,
+        missing_keys,
+        unexpected_keys,
+        error_msgs,
+    ):
+        old_prefix = prefix + "module."
+        for k in list(state_dict.keys()):
+            if k.startswith(old_prefix):
+                new_k = k.replace(old_prefix, prefix)
+                state_dict[new_k] = state_dict.pop(k)
+
+        super()._load_from_state_dict(
+            state_dict,
+            prefix,
+            local_metadata,
+            strict,
+            missing_keys,
+            unexpected_keys,
+            error_msgs,
+        )
+
     def forward(self, image):
         i2 = self.lc(image)
         i3 = nn.functional.relu(i2)
--- a/mmf/utils/checkpoint.py
+++ b/mmf/utils/checkpoint.py
@@ -321,11 +321,12 @@ class Checkpoint:
             key += "."
             value += "."
             for attr in ckpt:
+                if hasattr(model, "format_state_key"):
+                    formatted_attr = model.format_state_key(attr)
+                else:
+                    formatted_attr = attr
+
                 for own_attr in own_state:
-                    if hasattr(model, "format_state_key"):
-                        formatted_attr = model.format_state_key(attr)
-                    else:
-                        formatted_attr = attr
                     if (
                         key in own_attr
                         and value in formatted_attr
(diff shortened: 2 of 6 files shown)


mmf/modules/encoders.py
```diff
@@ -125,6 +125,15 @@ def __init__(self, config: Config, *args, **kwargs):
         self.lc.bias.data.copy_(torch.from_numpy(bias))
         self.out_dim = out_dim
 
+        def state_dict_key_override(state_dict, prefix, *args, **kwargs):
+            old_prefix = prefix + "module."
```
vedanuj, line 129: We shouldn't hardcode this IMO. In case there are other prefixes in future that we want to replace how can we handle it?

ytsheng: Right now, whenever there is code changes, we have to go into the top level format_state_key to perform state_key replacement. https://github.com/facebookresearch/mmf/search?q=format_state_key

Similarly to how we did it before, we'd have to modify the code to add additional prefixes that we wish to override.

mmf/utils/checkpoint.py
```diff
@@ -321,11 +321,12 @@ def _load_pretrained(self, ckpt):
             key += "."
             value += "."
             for attr in ckpt:
+                if hasattr(model, "format_state_key"):
+                    formatted_attr = model.format_state_key(attr)
+                else:
+                    formatted_attr = attr
+
```
apsdehal, line 328: Rebase on master.

mmf/modules/encoders.py
```diff
@@ -125,6 +125,15 @@ def __init__(self, config: Config, *args, **kwargs):
         self.lc.bias.data.copy_(torch.from_numpy(bias))
         self.out_dim = out_dim
 
+        def state_dict_key_override(state_dict, prefix, *args, **kwargs):
```
apsdehal, line 128: move this outside of the class. I am wondering if there is a better way to achieve this hooking via registry.

apsdehal: I have a feeling that directly overridding the `_load_from_state_dict` method might be more sustainable. See how pytorch does this: https://github.com/pytorch/pytorch/blob/8aaca4b46ad5856219a72c7164f67ab7211f70ab/torch/nn/quantized/modules/linear.py#L206

apsdehal: Also, check better way of doing this through class: https://github.com/pytorch/pytorch/blob/6dc763df3098485f158eeed2b9621cffaa229669/torch/nn/utils/spectral_norm.py#L149

apsdehal (changes requested):

mmf/utils/checkpoint.py
ytsheng: this is a minor refactor to move the formatted_attr assignment to the outer loop since it does not need to be computed every time.

mmf/modules/encoders.py
ytsheng: I changed it to using the `_load_from_state_dict to` override the state key

vedanuj (approved):

mmf/modules/encoders.py
```diff
@@ -125,6 +125,32 @@ def __init__(self, config: Config, *args, **kwargs):
...
+        unexpected_keys,
+        error_msgs,
+    ):
+        old_prefix = prefix + "module."
+        for k in list(state_dict.keys()):
+            if k.startswith(old_prefix):
+                new_k = k.replace(old_prefix, prefix)
+                state_dict[new_k] = state_dict.pop(k)
+
+        super(FinetuneFasterRcnnFpnFc7, self)._load_from_state_dict(
```
vedanuj, line 144: Change to Python3 parameterless super() call?

PR #677, opened 2020-11-06
[feat] allow passing in model's file path for from_pretrained

ytsheng:
- from_pretrained to allow a filepath
- add test

### DIFF
--- a/mmf/models/base_model.py
+++ b/mmf/models/base_model.py
@@ -52,6 +52,7 @@ from mmf.common.sample import to_device
 from mmf.modules.losses import Losses
 from mmf.utils.checkpoint import load_pretrained_model
 from mmf.utils.download import download_pretrained_model
+from mmf.utils.file_io import PathManager
 from omegaconf import MISSING, DictConfig, OmegaConf
 from torch import nn
 
@@ -211,15 +212,16 @@ class BaseModel(nn.Module):
         return results
 
     @classmethod
-    def from_pretrained(cls, model_name, *args, **kwargs):
-        model_key = model_name.split(".")[0]
-        model_cls = registry.get_model_class(model_key)
-        assert (
-            model_cls == cls
-        ), f"Incorrect pretrained model key {model_name} for class {cls.__name__}"
-        output = load_pretrained_model(model_name, *args, **kwargs)
+    def from_pretrained(cls, model_name_or_path, *args, **kwargs):
+        if not PathManager.isfile(model_name_or_path):
+            model_key = model_name_or_path.split(".")[0]
+            model_cls = registry.get_model_class(model_key)
+            assert (
+                model_cls == cls
+            ), f"Incorrect pretrained model key {model_name_or_path} "
+            "for class {cls.__name__}"
+        output = load_pretrained_model(model_name_or_path, *args, **kwargs)
         config, checkpoint = output["config"], output["checkpoint"]
-
         # Some models need registry updates to be load pretrained model
         # If they have this method, call it so they can update accordingly
         if hasattr(cls, "update_registry_for_pretrained"):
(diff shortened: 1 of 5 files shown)


mmf/models/base_model.py
```diff
@@ -213,9 +213,13 @@ def from_pretrained(cls, model_name, *args, **kwargs):
         assert (
             model_cls == cls
         ), f"Incorrect pretrained model key {model_name} for class {cls.__name__}"
-        output = load_pretrained_model(model_name, *args, **kwargs)
-        config, checkpoint = output["config"], output["checkpoint"]
 
+        if "from_file" in kwargs:
+            output = load_pretrained_model(kwargs["from_file"], *args, **kwargs)
+        else:
+            output = load_pretrained_model(model_name, *args, **kwargs)
```
vedanuj, line 220: The argument `model_name` should be able to take a file path, something like `model_name_or_file_path`. That will be more convenient to use and also similar to how other frameworks handle loading pretrained models. 

https://github.com/huggingface/transformers/blob/master/src/transformers/modeling_utils.py#L743

ytsheng: https://github.com/facebookresearch/mmf/blob/master/mmf/models/mmbt.py#L617

It looks like because of the above code, we'd need to specify what the model name is here rather than overriding the filepath together with the model. 

On the other hand, I think we should not have hardcoding the model in code like this, perhaps we should add the flag into the model config? @apsdehal

vedanuj: I agree it seems a bit hacky right now with hardcoded model name. I think in case of MMBT HM interface, we can pass a kwarg like "model_interface=True" that will return a `MMBTGridHMInterface` object for this specific case.

apsdehal: I think that use case is fine to ignore. If user is passing the file path directly one can return back the model in else condition. Please fix that in PR directly. Otherwise, Vedanuj said is important and should be implemented that way. Also, this usually is folder path not a file path where the checkpoint and config is stored. Can you fix `load_pretrained_model` so that it can directly take in a checkpoint and expect the checkpoint to have "config" attribute? Check `load_pretrained_model`'s implementation for more details.

apsdehal (changes requested):

mmf/models/base_model.py
ytsheng: Hi @apsdehal and @vedanuj, I updated the PR, let me know if you have more questions. Thanks for the feedback!

mmf/utils/checkpoint.py
```diff
@@ -34,8 +34,39 @@ def _hack_imports():
     )
 
 
-def load_pretrained_model(model_name_or_path, *args, **kwargs):
-    # If this is a file, then load this directly else download and load
+def _allowed_checkpoint_exts():
```
apsdehal, line 37: Make this a constant variable.

tests/models/interfaces/test_interfaces.py
```diff
@@ -13,10 +19,22 @@ class TestModelInterfaces(unittest.TestCase):
...
+        self._test_model_performance(model)
+
+    @test_utils.skip_if_no_network
+    @test_utils.skip_if_windows
+    @test_utils.skip_if_macos
+    def test_mmbt_hm_interface_from_file(self):
+        with tempfile.NamedTemporaryFile(suffix=".pth") as tmp:
+            self._create_checkpoint_file(tmp.name)
+
+            model = MMBT.from_pretrained(tmp.name, model_interface=True)
```
apsdehal, line 31: This is a non-backwards compatible change, we need to discuss this. In MMBT, can you specifically set `model_interface` to `True` by default?

mmf/models/mmbt.py
```diff
@@ -614,8 +614,9 @@ def from_pretrained(cls, model_name, *args, **kwargs):
         model = super().from_pretrained(model_name, *args, **kwargs)
         config = load_pretrained_model(model_name)["full_config"]
         OmegaConf.set_struct(config, True)
-        if model_name == "mmbt.hateful_memes.images":
+        if model_name == "mmbt.hateful_memes.images" or kwargs.get("model_interface"):
```
apsdehal, line 617: `interface` might be a better choice

apsdehal (approved):
Thanks for this. Please review the minor comments and feel free to land after that (After importing to phabricator)

PR #684, opened 2020-11-16
[fix] Fix test_nucleus_sampling after pytorch multinomial fix

BIT-silence:
Summary: Fix test_nucleus_sampling after pytorch multinomial fix

Differential Revision: D24997596


vedanuj (approved):
Thanks for fixing this!

PR #693, opened 2020-11-21
Changes to support training on tpus

ultrons:
@apsdehal created this PR as discussed. This seems to working reasonably with mmf_transformer training. Please review.

### DIFF
--- a/mmf/common/sample.py
+++ b/mmf/common/sample.py
@@ -396,7 +396,11 @@ def to_device(
     if isinstance(device, str):
         device = torch.device(device)
 
-    if not torch.cuda.is_available():
+    # default valude of device_type is cuda
+    # since other device types such as xla can be passed
+    # falling back to cpu should only happen when device_type
+    # is set to cude but cuda is not available.
+    if not torch.cuda.is_available() and device == "cuda":
         device = torch.device("cpu")
     # to_device is specifically for SampleList
     # if user is passing something custom built
--- a/mmf_cli/run.py
+++ b/mmf_cli/run.py
@@ -9,7 +9,7 @@ import torch
 from mmf.common.registry import registry
 from mmf.utils.build import build_config, build_trainer
 from mmf.utils.configuration import Configuration
-from mmf.utils.distributed import distributed_init, get_rank, infer_init_method
+from mmf.utils.distributed import distributed_init, get_rank, infer_init_method, is_xla
 from mmf.utils.env import set_seed, setup_imports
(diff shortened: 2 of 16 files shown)


mmf/common/sample.py
```diff
@@ -396,7 +396,11 @@ def to_device(
     if isinstance(device, str):
         device = torch.device(device)
 
-    if not torch.cuda.is_available():
+    # default valude of device_type is cuda
+    # since other device types such as xla can be passed
+    # falling back to cpu should only happen when device_type
+    # is set to cude but cuda is not available.
+    if not torch.cuda.is_available() and device == "cuda":
         device = torch.device("cpu")
```
ytsheng, line 404: do you mind adding a warning and/or log here?

ultrons: Thanks @ytsheng , I also identified other changes. I plan to push an update later this week which should be more review ready.

mmf_cli/run.py
```diff
@@ -108,15 +110,25 @@ def run(opts: typing.Optional[typing.List[str]] = None, predict: bool = False):
...
-        port = random.randint(10000, 20000)
-        config.distributed.init_method = f"tcp://localhost:{port}"
-        config.distributed.rank = None
-        torch.multiprocessing.spawn(
-            fn=distributed_main,
-            args=(configuration, predict),
-            nprocs=config.distributed.world_size,
-        )
+        if is_xla():
+            import torch_xla.distributed.xla_multiprocessing as xmp
```
apsdehal, line 114: Instead of polluting this file, it would be better to refactor these functions into a separate utils file for xla and then directly call them here for the sake of readability.

mmf_cli/run.py
```diff
@@ -108,15 +110,25 @@ def run(opts: typing.Optional[typing.List[str]] = None, predict: bool = False):
...
+            torch.multiprocessing.set_sharing_strategy("file_system")
+            xmp.spawn(
+                fn=distributed_main,
+                args=(configuration, predict),
+                nprocs=8,  # use all 8 TPU cores
```
apsdehal, line 119: Should this be configurable?

mmf_cli/run.py
```diff
@@ -108,15 +110,25 @@ def run(opts: typing.Optional[typing.List[str]] = None, predict: bool = False):
...
+                start_method='fork'
```
apsdehal, line 120: Should this also be configurable?

PR #700, opened 2020-11-28
[feat] D2 backbone intergation, Movie MCAN e2e

vedanuj:
[WIP] E2E with grid feats D2 models

### DIFF
--- a/mmf/datasets/processors/image_processors.py
+++ b/mmf/datasets/processors/image_processors.py
@@ -1,6 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 import collections
+import math
+import random
+import warnings
 
 import torch
 from mmf.common.registry import registry
@@ -82,3 +85,72 @@ class GrayScaleTo3Channels(BaseProcessor):
         if x.size(0) == 1:
             x = torch.cat([x] * 3, dim=0)
         return x
+
+
+@registry.register_processor("ResizeShortest")
+class ResizeShortest(BaseProcessor):
+    def __init__(self, *args, **kwargs):
+        min_size = kwargs["min_size"]
+        max_size = kwargs["max_size"]
+        if not isinstance(min_size, (list, tuple)):
+            min_size = (min_size,)
+        self.min_size = min_size
+        self.max_size = max_size
+
+    def get_size(self, image_size):
+        w, h = image_size
+        size = random.choice(self.min_size)
+        max_size = self.max_size
+        if max_size is not None:
+            min_original_size = float(min((w, h)))
+            max_original_size = float(max((w, h)))
+            if max_original_size / min_original_size * size > max_size:
+                size = int(math.floor(max_size * min_original_size / max_original_size))
+
+        if (w <= h and w == size) or (h <= w and h == size):
+            return (h, w)
+
+        if w < h:
+            ow = size
+            oh = int(size * h / w)
+        else:
+            oh = size
+            ow = int(size * w / h)
+
+        return (oh, ow)
+
+    def __call__(self, image):
+        size = self.get_size(image.size)
+        image = transforms.functional.resize(image, size)
+
+        return image
+
+
+@registry.register_processor("NormalizeBGR255")
+class NormalizeBGR255(BaseProcessor):
+    def __init__(self, *args, **kwargs):
+        self.mean = kwargs["mean"]
+        self.std = kwargs["std"]
+        self.to_bgr255 = kwargs["to_bgr255"]
+        self.pad_size = kwargs["pad_size"]
+        if self.pad_size > 0:
+            warnings.warn(
+                f"You are setting pad_size > 0, tensor will be padded to a fix size of"
+                f"{self.pad_size}. "
+                f"The image_mask will cover the pad_size of {self.pad_size} instead of"
+                "the original size."
+            )
+
+    def __call__(self, image):
+        if self.to_bgr255:
+            image = image[[2, 1, 0]] * 255
+        image = transforms.functional.normalize(image, mean=self.mean, std=self.std)
+        if self.pad_size > 0:
+            assert (
+                self.pad_size >= image.shape[1] and self.pad_size >= image.shape[2]
+            ), f"image size: {image.shape}"
+            padded_image = image.new_zeros(3, self.pad_size, self.pad_size)
+            padded_image[:, : image.shape[1], : image.shape[2]] = image.clone()
+
+            return padded_image
+        return image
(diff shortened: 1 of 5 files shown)


mmf/modules/encoders.py
```diff
@@ -22,6 +24,12 @@
 from transformers.modeling_auto import AutoModel
 
 
+try:
+    from detectron2.modeling import ShapeSpec, build_resnet_backbone
+except ImportError:
+    pass
```
apsdehal, line 30: It would make more sense to import it below where it is actually required. In case it is missing then, it would be to tell the user that detectron2 needs to be installed before using that particular resnet encoder.

mmf/modules/encoders.py
```diff
@@ -230,6 +241,50 @@ def forward(self, x):
         return out  # BxNx2048
 
 
+@registry.register_encoder("resnet")
+class ResNetEncoder(Encoder):
```
apsdehal, line 245: @kdexd 's PR #336 does have a more generic version of this. Not sure what his current plans are regarding landing this.

projects/movie_mcan/configs/vqa2/defaults.yaml
```diff
@@ -12,13 +12,13 @@ dataset_config:
       - vqa2.grid_features
     features:
       train:
-      - vqa2/grid_features/features
-      - vqa2/grid_features/features
-      - vqa2/grid_features/features
+      - /checkpoint/vedanuj/features
```
apsdehal, line 15: Do they need to be updated?

projects/movie_mcan/configs/vqa2/e2e.yaml
```diff
@@ -0,0 +1,93 @@
...
+- ./defaults.yaml
+
+model_config:
+  movie_mcan:
+    image_feature_encodings:
+      type: resnet
+      params:
+        backbone: d2_resnet
+        pretrained: true
+        pretrained_path: /checkpoint/vedanuj/X-152pp.pth
```
apsdehal, line 11: This needs to be uploaded somewhere. Detectron2 function probably supports URL.

projects/movie_mcan/configs/vqa2/e2e.yaml
```diff
@@ -0,0 +1,93 @@
...
+dataset_config:
+  vqa2:
+    use_images: true
+    use_features: false
+    zoo_requirements:
+      - vqa2.defaults
+      - vqa2.grid_features
+    images:
+      train:
+      - /private/home/vedanuj/.cache/torch/mmf/data/datasets/coco/defaults/images
```
apsdehal, line 61: Update this to MMF paths.

mmf/modules/encoders.py
```diff
@@ -230,6 +241,50 @@ def forward(self, x):
...
+    @dataclass
+    class Config(Encoder.Config):
+        name: str = "resnet"
+        backbone: str = None
+        pretrained: bool = True
+        pretrained_path: str = None
+
+    def __init__(self, config: Config, *args, **kwargs):
```
apsdehal, line 253: If we can write simple tests to test the loading and output that would be great to make sure this doesn't break in future.

mmf/datasets/processors/image_processors.py
```diff
@@ -82,3 +85,74 @@ def transform(self, x):
...
+@registry.register_processor("NormalizeBGR255")
+class NormalizeBGR255(BaseProcessor):
+    def __init__(self, *args, **kwargs):
+        self.mean = kwargs["mean"]
+        self.std = kwargs["std"]
+        self.to_bgr255 = kwargs["to_bgr255"]
+        self.pad_size = kwargs["pad_size"]
+        if self.pad_size > 0:
+            warnings.warn(
+                "Padding is automatically supported in SampleList to the maximum size "
```
apsdehal, line 138: Did we add this to SampleList, I don't remember it from the top of my head.

mmf/models/movie_mcan.py
```diff
@@ -80,7 +80,7 @@ def _init_feature_encoders(self, attr: str):
         with omegaconf.open_dict(feat_encoder_config):
             feat_encoder_config.params.model_data_dir = self.config.model_data_dir
             feat_encoder_config.params.in_dim = feature_dim
-        feat_model = build_image_encoder(feat_encoder_config, direct_features=True)
+        feat_model = build_image_encoder(feat_encoder_config, direct_features=False)
```
apsdehal, line 83: Will this still support the older encoder? Otherwise, it would make sense to rely on `build_encoder` function.

mmf/models/movie_mcan.py
```diff
@@ -226,7 +226,13 @@ def process_feature_embedding(
             feature = sample_list.image
 
             feature_encoder = getattr(self, attr + "_feature_encoders")
-            encoded_feature = feature_encoder(feature, text_embedding_vec)
```
apsdehal, line 229: Just for my knowledge, does removing the `text_embedding_vec` change anything. I checked that the default config is only default encoder. Maybe I am misunderstanding something here.

apsdehal (changes requested):
Thanks for implementing this. I have tested it offline and it works. I have left some comments which once addressed we can go ahead and land this.

mmf/modules/encoders.py
kdexd: #336 has some merge conflicts, and I am still revamping it with new upstream changes. 
However, image encoder seems like an important feature that would be nice to have in upstream ASAP.

I scoped it out separately and made a quick short PR: https://github.com/facebookresearch/mmf/pull/728
Once this is merged, it could be used in this PR if serves the purpose (I will re-use it in my PR).

mmf/models/movie_mcan.py
vedanuj: There was no old encoder. It was using the feature from sample list directly. This won't change that behavior.

mmf/modules/encoders.py
vedanuj: I had tried that earlier. Moving it inside the encoder class causes an error in detectron2's call to fvcore, where it tries to jit compile an unrelated `sigmoid_focal_loss` function and throws error. I will need more time to investigate why that is occuring.

```
torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults
```

mmf/modules/encoders.py
vedanuj: I have removed the support for torchvision resnets in this method and made it specific to detectron2 resnet backbones only. @kdexd's implementation won't overlap with this anymore.

mmf/models/movie_mcan.py
vedanuj: `text_embedding_vec` is not required for image feature encoders. This was old code and was not verified properly in MMF.

apsdehal (approved):
Super! Please add tests for the encoders and processors and docs if there in a separate PR.

PR #715, opened 2020-12-15
[fix] fix typo in model zoo config

ytsheng:


vedanuj (approved):
Thanks for fixing this.

PR #716, opened 2020-12-16
[WIP][feat] per-dataset evaluation loop w/ dataset-level metrics

ronghanghu:
This is the first piece of the PRs for the multimodal multitask transformer merge.

------

Summary
- change the current evaluation loop (that runs on the union of all datasets) to per-dataset evaluation.
- generate prediction JSON object during evaluation loop, which can be consumed by metrics that evaluate the entire dataset (e.g. mAP for object detection or CIDEr for image captioning) and cannot be expressed as averaging over per-batch metrics

Issues
- in the current evaluation, the `total_loss` results will only retain the `total_loss` metric from the last dataset. Shall we give `<dataset_name>/total_loss` instead?

### DIFF
--- a/mmf/common/test_reporter.py
+++ b/mmf/common/test_reporter.py
@@ -56,9 +56,12 @@ class TestReporter(Dataset):
 
         PathManager.mkdirs(self.report_folder)
 
-    def next_dataset(self):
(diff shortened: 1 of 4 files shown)


ronghanghu:
This PR is superseded by #733

PR #728, opened 2021-01-10
[feat] Add generic TorchvisionResnetImageEncoder in ImageEncoderFactory.

kdexd:
This PR adds a generic image encoder named `TorchvisionResnetImageEncoder` in `mmf.modules.ImageEncoderFactory` — this encoder can instantiate any ResNet-like model from [`torchvision.models`](https://pytorch.org/docs/stable/torchvision/models.html) based on its name. The encoder returns spatial grid features same as torchvision.

It can be instantiated using a minimal config as:

```yaml
type: "torchvision_resnet"
params:
    name: <name>
    pretrained: false
```

- Set `pretrained: true` for loading ImageNet-pretrained weights from torchvision.
- Supported `<name>`s are:
    - `resnet18`, `resnet34`, `resnet50`, `resnet101`, `resnet152`
    - `resnext50_32x4d`, `resnext101_32x8d`
    - `wide_resnet50_2`, `wide_resnet101_2`

### Example

```python
>>> import torch
>>> from omegaconf import OmegaConf
>>> from mmf.utils.build import build_encoder
>>> 
>>> config = OmegaConf.load("encoder_config.yaml")
>>> encoder = build_encoder(config)
>>> images = torch.randn(32, 3, 224, 224)   # shape: (b, c, h, w)
>>> grid_feats = encoder(images)  # shape: (b, 2048, 7, 7)
```

### Backward Compatibility

This PR maintains backward compatibility as it adds a new encoder without changing existing ones. However, it subsumes the functionality of [ResNet152ImageEncoder](https://github.com/facebookresearch/mmf/blob/d04914a7bbae28f33356e8b17113a81d6291a5dc/mmf/modules/encoders.py#L215), hence the latter could be later replaced and deprecated in favor of this addition.

### DIFF
--- a/mmf/modules/encoders.py
+++ b/mmf/modules/encoders.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from copy import deepcopy
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any
+from typing import Any, Dict
 
 import torch
 import torchvision
@@ -185,6 +185,7 @@ class IdentityEncoder(Encoder):
 class ImageEncoderTypes(Enum):
     default = "default"
     identity = "identity"
+    torchvision_resnet = "torchvision_resnet"
     resnet152 = "resnet152"
     detectron2_resnet = "detectron2_resnet"
 
@@ -208,9 +209,10 @@ class ImageEncoderFactory(EncoderFactory):
             self.module.out_dim = params.in_dim
         elif self._type == "resnet152":
             self.module = ResNet152ImageEncoder(params)
+        elif self._type == "torchvision_resnet":
+            self.module = TorchvisionResNetImageEncoder(params)
         elif self._type == "detectron2_resnet":
             self.module = Detectron2ResnetImageEncoder(params)
-
         else:
             raise NotImplementedError("Unknown Image Encoder: %s" % self._type)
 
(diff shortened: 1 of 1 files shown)


kdexd:
Additional Note: This encoder will not support other torchvision models like VGG, GoogleNet, ShuffleNet etc. — these classes have a different module naming convention and different output feature dimensions. Supporting all models in a single class may get too cluttered. We can introduce the functionality in a separate encoder for these model families if needed.

vedanuj:
This looks good. Can you please rebase the PR?

kdexd:
@vedanuj: Sounds good. I had one thought about the compatibility with the rest of MMF codebase. Do you think it may be useful to add a config parameter to control the format of output features — `"grid"` (BCHW, b x 2048 x 7 x 7), `"pooled"` (b x 2048) or `"vectors"` (b x 49 x 2048).

vedanuj:
> @vedanuj: Sounds good. I had one thought about the compatibility with the rest of MMF codebase. Do you think it may be useful to add a config parameter to control the format of output features — `"grid"` (BCHW, b x 2048 x 7 x 7), `"pooled"` (b x 2048) or `"vectors"` (b x 49 x 2048).

I think this format is fine. I would leave it to the model to decide what it wants to do with the output from the encoder. I know few existing encoders handle the output format, but I feel that is not the right design. In the long run it will lead to  encoders getting bloated with model specific output format handling logic. Let me know if you think otherwise.

kdexd:
Makes sense, then this PR is finished from my end!

PR #733, opened 2021-01-15
[WIP][feat] Per dataset evaluation loop and metric; detection datasets; UniT model

ronghanghu:
Summary
- Per dataset evaluation loop and metric
  - change the current evaluation loop (that runs on the union of all datasets) to per-dataset evaluation.
  - allow running metrics only on a subset with `datasets` config field (see example below)
  - generate prediction JSON object during evaluation loop, which can be consumed by metrics that evaluate the entire dataset (e.g. mAP for object detection or CIDEr for image captioning) and cannot be expressed as averaging over per-batch metrics
 - detection dataset support
   - COCO detection dataset
   - Visual Genome detection (object and attribute) dataset
   - relevant parts (e.g. package dependencies, test reporter, batch collator, etc.)
 - UniT model
   - (moved several hard-coded components into config, such as MT_DATASETS)
   - tested w/ previous snapshots -- results match the paper on COCO, VG and VQAv2

Example on specifying per-dataset metric with `datasets` (here `vqa_accuracy` will only run on vqa2 while `detection_mean_ap` will only run on detection_coco and detection_visual_genome). If `datasets` is not specified, the default behavior is to run on all dataset:
```
evaluation:
  metrics:
  - type: vqa_accuracy
    datasets:
    - vqa2
  - type: detection_mean_ap
    datasets:
    - detection_coco
    - detection_visual_genome
 ```

### DIFF
--- a/mmf/common/batch_collator.py
+++ b/mmf/common/batch_collator.py
@@ -21,7 +21,7 @@ class BatchCollator:
         ):
             sample_list = batch[0]
         elif not isinstance(batch, SampleList):
-            sample_list = SampleList(batch)
+            sample_list = _build_sample_list_with_detr_fields(batch)
 
         if sample_list._get_tensor_field() is None:
             sample_list = SampleList(sample_list.to_dict())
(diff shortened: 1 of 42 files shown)


ronghanghu:
(it seems to run into some weird isort issue that I couldn't reproduce locally)

vedanuj:
Can this be closed in favor of #739 #740 #741 ?

PR #735, opened 2021-01-16
[fix] Allow loading checkpoint from a folder

apsdehal:
- Needed for e2e movie+mcan

Test Plan:

- Tested locally with e2e movie+mcan handler
- Adds tests for loading from a folder
- Combines older file based test to main pretrained as they had
sequential dep

### DIFF
--- a/mmf/models/base_model.py
+++ b/mmf/models/base_model.py
@@ -213,7 +213,9 @@ class BaseModel(nn.Module):
 
     @classmethod
     def from_pretrained(cls, model_name_or_path, *args, **kwargs):
-        if not PathManager.isfile(model_name_or_path):
+        # Check if the path exists, if not it is pretrained, otherwise,
+        # we will try to load the checkpoint from the path
+        if not PathManager.exists(model_name_or_path):
             model_key = model_name_or_path.split(".")[0]
             model_cls = registry.get_model_class(model_key)
             assert (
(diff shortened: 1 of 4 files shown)


mmf/models/base_model.py
```diff
@@ -213,15 +213,27 @@ def format_for_prediction(self, results, report):
...
+        config, checkpoint, full_config = (
+            output["config"],
+            output["checkpoint"],
+            output["full_config"],
+        )
+
+        # Save original config for state reset later
+        config_temp_holder = registry.get("config")
+        # Register full config from checkpoint when loading the model
+        registry.register("config", full_config)
```
ytsheng, line 235: Thanks for fixing! Should we register another param in the registry `model_config` that is different so that not to override the same thing which might cause externalities?

mmf/models/movie_mcan.py
```diff
@@ -24,7 +24,9 @@ def __init__(self, config):
         super().__init__(config)
         self.config = config
         self._global_config = registry.get("config")
```
ytsheng, line 26: I would prefer global_config just be a variable rather than an instance variable (not owned by self). Because I really don't see it being used anywhere else. https://fburl.com/k0qak2bp

mmf/models/movie_mcan.py
```diff
@@ -24,7 +24,9 @@ def __init__(self, config):
...
-        self._datasets = self._global_config.datasets.split(",")
+        self._datasets = self._global_config.datasets
+        if isinstance(self._datasets, str):
```
ytsheng, line 28: what is the alternative, should we assert/warning there?

ytsheng (approved):
Looks good

mmf/models/base_model.py
apsdehal: IMO, we should just keep one config in registry that is the global config to avoid any confusions. `model_config` can anyways be accessed from global config.

mmf/models/movie_mcan.py
apsdehal: An alternative is not really needed, this usually happens when the config is loaded from a file or checkpoint. So, this is just handling that edge case.

mmf/models/movie_mcan.py
apsdehal: Let's handle that in a separate PR. To unblock intern, let's first land this.

vedanuj (approved):
Looks good to me!

PR #736, opened 2021-01-18
[feat] GLUE datasets from Hugginface datasets library

apsdehal:
- Uses single interface to load 4 GLUE Datasets
- Currently, only adds 4 datasets needed by UniT, can be extended easily
to more
- Specify as `glue_qnli`, `glue_mnli_mismatched` etc.

Test Plan:

Tested locally and verified the outputs

### DIFF
--- a/mmf/datasets/builders/glue/builder.py
+++ b/mmf/datasets/builders/glue/builder.py
@@ -1,0 +1,97 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+import logging
+
+import torch
+from mmf.common.registry import registry
+from mmf.common.sample import Sample
+from mmf.datasets.base_dataset import BaseDataset
+from mmf.datasets.mmf_dataset_builder import MMFDatasetBuilder
+from mmf.utils.general import retry_n
+
+
+MAX_RETRIES = 9
+
+logger = logging.getLogger()
+
+
+class GLUEDataset(BaseDataset):
+    DATASET_KEY_MAP = {
+        "text_a": {
+            "glue_mnli_mismatched": "premise",
+            "glue_qnli": "question",
+            "glue_sst2": "sentence",
+            "glue_qqp": "question1",
+        },
+        "text_b": {
+            "glue_mnli_mismatched": "hypothesis",
+            "glue_qnli": "sentence",
+            "glue_sst2": None,
+            "glue_qqp": "question2",
+        },
+    }
+
+    def __init__(self, config, dataset_type, imdb_idx):
+        try:
+            from datasets import load_dataset
+        except ModuleNotFoundError:
+            logger.error(
+                "Please install 'datasets' library by running `pip install datasets`."
+            )
+            raise
+
+        dataset_name = f"glue_{config.task}"
+        super().__init__(dataset_name, config, dataset_type, imdb_idx)
+
+        if dataset_type == "val":
+            # datasets library uses validation as val set.
+            dataset_type = "validation"
+
+        # For MNLI-MM, set train to MNLI-M train
+        task = config.task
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 torch==1.6.0
 torchvision==0.7.0
 numpy>=1.16.6
-tqdm>=4.43.0
+tqdm>=4.43.0,<4.50.0
```
ronghanghu, line 4: `tqdm` has to be lower than 4.50.0 -- otherwise incompatible with `datasets==1.2.1` (already the latest).

apsdehal: I will just try to upgrade it to 4.50.0.

mmf/datasets/builders/glue/builder.py
```diff
@@ -0,0 +1,87 @@
...
+        return len(self.dataset)
+
+    def __getitem__(self, idx):
+        annotation = self.dataset[idx]
+        current_sample = Sample()
+        text_processor_input = {
+            "text_a": annotation[self.DATASET_KEY_MAP["text_a"][self.dataset_name]]
+        }
+
+        text_b = annotation.get(self.DATASET_KEY_MAP["text_b"][self.dataset_name], None)
```
ronghanghu, line 58: I added a fix-up commit for the case of `text_b` non-existence (this line is the new version). @apsdehal Could you import it again? Thanks 😃

mmf/datasets/builders/glue/builder.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@registry.register_builder("glue_sst2")
+@registry.register_builder("glue_mnli_mismatched")
+@registry.register_builder("glue_qqp")
+@registry.register_builder("glue_qnli")
+class GLUEBuilder(MMFDatasetBuilder):
+    def __init__(self, dataset_name="glue", dataset_class=GLUEDataset, *args, **kwargs):
+        super().__init__(dataset_name, dataset_class)
+        self.dataset_name = "glue"
```
ytsheng, line 74: self.dataset_name = dataset_name

ytsheng:
Looks good to me otherwise!

giphy

mmf/datasets/builders/glue/builder.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+import torch
+from mmf.common.registry import registry
+from mmf.common.sample import Sample
+from mmf.datasets.base_dataset import BaseDataset
+from mmf.datasets.mmf_dataset_builder import MMFDatasetBuilder
+from mmf.utils.general import retry_n
+
+from datasets import load_dataset
```
vedanuj, line 9: Is `datasets` supported internally in fbcode? Changes in TARGET file are missing if it is supported. Otherwise have to figure out a way to enable this only in OSS.

mmf/datasets/builders/glue/builder.py
```diff
@@ -0,0 +1,87 @@
...
+
+@registry.register_builder("glue_sst2")
+@registry.register_builder("glue_mnli_mismatched")
+@registry.register_builder("glue_qqp")
+@registry.register_builder("glue_qnli")
+class GLUEBuilder(MMFDatasetBuilder):
+    def __init__(self, dataset_name="glue", dataset_class=GLUEDataset, *args, **kwargs):
+        super().__init__(dataset_name, dataset_class)
+        self.dataset_name = "glue"
+        self.set_dataset_class(GLUEDataset)
```
vedanuj, line 75:
```suggestion
        self.set_dataset_class(dataset_class)
```

PR #738, opened 2021-01-18
[enhancement] be more explicit in isort

ytsheng:
* print in ci the difference that isort would make
* allow isort to be more explicit.
* my pytorch lightning PR is experiencing some local/circleci inconsistencies for isort despite using the same version.


apsdehal (approved):
Super! This would be really helpful!

PR #739, opened 2021-01-19
[feat] multitask support: per-dataset evaluation loop, metrics, and sampling ratios

ronghanghu:
This PR is built upon https://github.com/facebookresearch/mmf/pull/736

- change the current evaluation loop (that runs on the union of all
datasets) to per-dataset evaluation.
- generate prediction JSON object during evaluation loop, which can be
consumed by metrics that evaluate the entire dataset (e.g. mAP for
object detection or CIDEr for image captioning) and cannot be expressed
as averaging over per-batch metrics
- run metrics on a subset of datasets with (optional) `datasets` config
- allow specifying per-dataset sampling ratio for multi-task training

Test Plan:

Tested locally and verified the outputs

---
Example on specifying per-dataset metric with `datasets` (here `vqa_accuracy` will only run on vqa2 while `detection_mean_ap` will only run on detection_coco and detection_visual_genome). If `datasets` is not specified, the default behavior is to run on all dataset:
```
evaluation:
  metrics:
  - type: vqa_accuracy
    datasets:
    - vqa2
  - type: detection_mean_ap
    datasets:
    - detection_coco
    - detection_visual_genome
 ```
 
---
 Examples specifying per-dataset sampling ratio during training through `multitasking.enabled` and `multitasking.sampling_ratios`:
 ```
 multitasking:
  enabled: true
  sampling_ratios:
    detection_coco: 0.2
    detection_visual_genome: 0.07
    visual_entailment: 0.12
    vqa2: 0.26
    glue_qnli: 0.1
    glue_mnli_mismatched: 0.1
    glue_qqp: 0.1
    glue_sst2: 0.05
 ```

### DIFF
--- a/mmf/common/test_reporter.py
+++ b/mmf/common/test_reporter.py
@@ -56,9 +56,12 @@ class TestReporter(Dataset):
 
         PathManager.mkdirs(self.report_folder)
 
-    def next_dataset(self):
+    def next_dataset(self, flush_report=True):
         if self.current_dataset_idx >= 0:
-            self.flush_report()
+            if flush_report:
+                self.flush_report()
+            else:
+                self.report = []
 
         self.current_dataset_idx += 1
 
@@ -95,6 +98,10 @@ class TestReporter(Dataset):
         logger.info(f"Wrote predictions for {name} to {os.path.abspath(filepath)}")
         self.report = []
 
+    def postprocess_dataset_report(self):
+        if hasattr(self.current_dataset, "on_prediction_end"):
+            self.report = self.current_dataset.on_prediction_end(self.report)
+
     def csv_dump(self, filepath):
         with PathManager.open(filepath, "w") as f:
             title = self.report[0].keys()
@@ -123,12 +130,12 @@ class TestReporter(Dataset):
     def __getitem__(self, idx):
         return self.current_dataset[idx]
 
-    def add_to_report(self, report, model):
+    def add_to_report(self, report, model, execute_on_master_only=True):
         keys = ["id", "question_id", "image_id", "context_tokens", "captions", "scores"]
         for key in keys:
             report = self.reshape_and_gather(report, key)
 
-        if not is_master():
+        if execute_on_master_only and not is_master():
             return
 
         results = self.current_dataset.format_for_prediction(report)
--- a/mmf/trainers/core/evaluation_loop.py
+++ b/mmf/trainers/core/evaluation_loop.py
@@ -17,34 +17,56 @@ logger = logging.getLogger(__name__)
 
 class TrainerEvaluationLoopMixin(ABC):
     def evaluation_loop(
-        self, loader, use_tqdm: bool = False, single_batch: bool = False
+        self, dataset_type: str, use_tqdm: bool = False, single_batch: bool = False
     ) -> Tuple[Dict[str, Any], Type[Meter]]:
         meter = Meter()
+        reporter = self.dataset_loader.get_test_reporter(dataset_type)
 
         with torch.no_grad():
             self.model.eval()
             disable_tqdm = not use_tqdm or not is_master()
-            combined_report = None
 
-            for batch in tqdm.tqdm(loader, disable=disable_tqdm):
-                report = self._forward(batch)
-                self.update_meter(report, meter)
+            while reporter.next_dataset(flush_report=False):
+                dataloader = reporter.get_dataloader()
 
-                # accumulate necessary params for metric calculation
-                if combined_report is None:
-                    combined_report = report
-                else:
-                    combined_report.accumulate_tensor_fields_and_loss(
-                        report, self.metrics.required_params
-                    )
-                    combined_report.batch_size += report.batch_size
+                combined_report = None
+                for batch in tqdm.tqdm(dataloader, disable=disable_tqdm):
+                    prepared_batch = reporter.prepare_batch(batch)
+                    model_output = self.model(prepared_batch)
+                    report = Report(prepared_batch, model_output)
 
-                if single_batch is True:
-                    break
+                    self.update_meter(report, meter)
 
-            combined_report.metrics = self.metrics(combined_report, combined_report)
-            self.update_meter(combined_report, meter, eval_mode=True)
-            logger.info("Validation Done")
+                    # accumulate necessary params for metric calculation
+                    if combined_report is None:
+                        # make a copy of report since `reporter.add_to_report` will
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: mmf/common/test_reporter.py [hunk 1]
```diff
@@ -95,6 +98,10 @@ def flush_report(self):
         logger.info(f"Wrote predictions for {name} to {os.path.abspath(filepath)}")
         self.report = []
 
+    def postprocess_dataset_report(self):
+        if hasattr(self.current_dataset, "on_prediction_end"):
+            self.report = self.current_dataset.on_prediction_end(self.report)
```
apsdehal, line 103: @vedanuj What do you think is the right way to do this? Should this be handled via callbacks?

mmf/trainers/core/evaluation_loop.py
```diff
@@ -17,33 +17,53 @@
 
 class TrainerEvaluationLoopMixin(ABC):
     def evaluation_loop(
-        self, loader, use_tqdm: bool = False, single_batch: bool = False
+        self, dataset_type: str, use_tqdm: bool = False, single_batch: bool = False
     ) -> Tuple[Dict[str, Any], Type[Meter]]:
         meter = Meter()
+        reporter = self.dataset_loader.get_test_reporter(dataset_type)
```
apsdehal, line 23: Since, after this change in `DatasetLoader`, val_dataset and test_dataset are essentially useless, I would suggest that we move this to `DatasetLoader` and instead of using `MulitDatasetLoader` to initialize val and test dataloader use this. Also, now, I think it is right time to move `test_reporter.py` to be under datasets folder as well.

mmf/trainers/core/evaluation_loop.py
```diff
@@ -17,33 +17,53 @@
...
-                report = self._forward(batch)
-                self.update_meter(report, meter)
+            while reporter.next_dataset(flush_report=False):
+                dataloader = reporter.get_dataloader()
+
+                combined_report = None
+                for batch in tqdm.tqdm(dataloader, disable=disable_tqdm):
+                    prepared_batch = reporter.prepare_batch(batch)
+                    model_output = self.model(prepared_batch)
+                    report = Report(prepared_batch, model_output)
```
apsdehal, line 36: Once the above mentioned change happens, please move back to using `self._forward`.

mmf/trainers/core/evaluation_loop.py
```diff
@@ -17,33 +17,53 @@
...
+                        combined_report.batch_size += report.batch_size
 
-            combined_report.metrics = self.metrics(combined_report, combined_report)
-            self.update_meter(combined_report, meter, eval_mode=True)
+                    # Each node generates a separate copy of predict JSON from the report,
+                    # which will be used to evaluate dataset-level metrics
+                    # (such as mAP in object detection or CIDEr in image captioning)
+                    # Since `reporter.add_to_report` changes report keys (e.g. scores),
+                    # do this after `combined_report.accumulate_tensor_fields`
+                    reporter.add_to_report(report, self.model, master_only=False)
```
apsdehal, line 56: Isn't this unnecessary calculation for dataset which don't actually output this as well as for metrics which don't need this?

apsdehal (changes requested):
Thanks for the change! I requested some changes before moving forward. @ytsheng can you take a look at this PR and check if changes are required on lightning side as well.

apsdehal:
Also, please check if tests need to be extended to test this change better.

mmf/trainers/core/evaluation_loop.py
ronghanghu: Yes, that's true. What do you feel is a better way? Shall I add "test_report" was a `required_params` for each metric, and only gather report if "test_report" is in `required_params`?

mmf/trainers/core/evaluation_loop.py
ronghanghu: After further inspection, I think it would be better to keep the `MulitDatasetLoader` for val and test datasets, for the following reasons:

1) val_dataset and test_dataset are still used in this case -- currently, `TestReporter` dependents on `MulitDatasetLoader` to load each evaluation dataset. During execution, it accesses each test dataset through `MulitDatasetLoader`'s `get_datasets` function.
2) It would be better to keep just one, same mechanism to load multiple datasets rather than maintaining a separate multi-dataset loading mechanism, so using those that already exists in `MulitDatasetLoader` probably works better.
3) Keeping `MulitDatasetLoader` for val and test datasets would lead to a smaller refactorization and more backward compatibility in the MMFTrainer (so that if people subclass this trainer and uses val datasets, it won't be broken).

ronghanghu:
@apsdehal I revised this PR, adding `__prediction_report__` in a metric's `required_params` so that prediction reports will only be generated when necessary. For val and test loader, I kept them as-is due to the reasons mentioned in the comments above.

mmf/trainers/core/evaluation_loop.py
ronghanghu: (Added `__prediction_report__` in a metric's `required_params`)

mmf/common/test_reporter.py
```diff
@@ -123,12 +130,12 @@ def __len__(self):
     def __getitem__(self, idx):
         return self.current_dataset[idx]
 
-    def add_to_report(self, report, model):
+    def add_to_report(self, report, model, master_only=True):
```
vedanuj, line 133: Users might find this name a bit confusing and how it differentiates with `is_master`. Can we make it more succinct  with something like `execute_on_master_only`? Thoughts?

mmf/common/test_reporter.py
vedanuj: The right way to do this will be to add a new Reporting callback, with hooks `on_validation_end` or `on_prediction_end` as required. However this might be fine for now given we are moving to lightning soon. 

cc @ytsheng Does Lightning have any callbacks for report generation or report gathering after validation/test?

mmf/trainers/core/evaluation_loop.py
apsdehal: I like your suggestion, but maintainability wise it is an extra burden. I will fix this in my change around datamodules.

apsdehal (approved):
This LGTM! Please ask @ytsheng about Vedanuj's comment and probably try to resolve it.

mmf/common/test_reporter.py
ronghanghu: (changed to execute_on_master_only)

Issue #725, opened 2021-01-02
Recall Metrics will result in Recursion error


hqsiswiliam:
## Command
Any running command
## To Reproduce

1. set model metrics to `r@5` (any recall metrics)
2. run command until validation run
3. There will be an error (*provided in Additional context section*)

## Expected behavior

Successfully validating at `r@5`

## Environment
Collecting environment information...
PyTorch version: 1.6.0
Is debug build: No
CUDA used to build PyTorch: 10.2

OS: Zorin OS 15.2
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
CMake version: version 3.10.2

Python version: 3.7
Is CUDA available: Yes
CUDA runtime version: 10.0.130
GPU models and configuration: GPU 0: GeForce RTX 2070
Nvidia driver version: 440.82
cuDNN version: /usr/local/cuda-10.1/targets/x86_64-linux/lib/libcudnn.so.7

Versions of relevant libraries:
[pip3] numpy==1.19.4
[pip3] torch==1.6.0
[pip3] torchtext==0.5.0
[pip3] torchvision==0.7.0
[conda] numpy                     1.19.4                   pypi_0    pypi
[conda] torch                     1.6.0                    pypi_0    pypi
[conda] torchtext                 0.5.0                    pypi_0    pypi
[conda] torchvision               0.7.0                    pypi_0    pypi

## Additional context

In `metric.py` file, we have
```python
@registry.register_metric("r@5")
class RecallAt5(RecallAtK):
    """
    Calculate Recall@5 which specifies how many time the chosen candidate
    was among first 5 rank.

    **Key**: ``r@5``.
    """

    def __init__(self):
        super().__init__("r@5")

    def calculate(self, sample_list, model_output, *args, **kwargs):
        """Calculate Recall@5 and return it back.

        Args:
            sample_list (SampleList): SampleList provided by DataLoader for
                                current iteration
            model_output (Dict): Dict returned by model.

        Returns:
            torch.FloatTensor: Recall@5

        """
        return self.calculate(sample_list, model_output, k=5)
```
But please notice that the `self.calculate` call will cause the infinite self-recursion on the method
The error is:
```
RecursionError: maximum recursion depth exceeded in comparison
```
---
Meanwhile, when I tried to do a monkeypatch on this method (I copied and renamed `calculate` in `RecallAtK`, and utilized the newly copied method in `RecallAt5`), I got an error:
```
[...]

PR #746, opened 2021-01-22
[fix, test] Fix recall metric bug, add test

brettallenyo:
- Fix infinite recursion error #725
- Add back mistakenly removed code
- Add test for recall metric at 1, 5, and 10

This is a response to bug #725 

I believe `process_ranks` may have been accidentally deleted in this commit so I added it back: 
https://github.com/facebookresearch/mmf/commit/6b6f4fc221c46f8c55d55a06e110f3d8f9ca2417

I also added a testing file, which basically ranks all of the expected values in increasing order (so the last candidate is always the chosen one).

Both the expected values and predicted values are (10, 100) tensors. As the index of top level array increases, I alter the final value to be decreasing by 2, so the chosen candidates are 198 (maximum value 99*2), 191, 189, 187, ...

I decrease it in this way so that each of the candidates become out of the bounds for the increasing Recall Metrics. The rankings are `tensor([ 1,  4,  5,  6,  7,  8,  9, 10, 11, 12])`

This makes the results varying for each metric:

`RecallAt1` --> 1/10 = .1
`RecallAt5` --> 3/10 = .3
`RecallAt10` --> 8/10 = .8

I would love for people to take a look at `metrics.py` line `462` (`process_ranks`)
There are print statements in here from the previous commit, but I do not see print statements anywhere else in the modules, should I delete these, leave them, do we have a different logging style? Thanks!

### DIFF
--- a/mmf/modules/metrics.py
+++ b/mmf/modules/metrics.py
@@ -45,6 +45,7 @@ Example config for above metric::
 """
 
 import collections
+import warnings
 
 import torch
 from mmf.common.registry import registry
@@ -459,6 +460,22 @@ class RecallAtK(BaseMetric):
             gt_ranks[i] = int(ranks[i, ans_ind[i].long()])
         return gt_ranks
 
+    def process_ranks(self, ranks):
+        num_opts = 100
+
+        # none of the values should be 0, there is gt in options
+        if torch.sum(ranks.le(0)) > 0:
+            num_zero = torch.sum(ranks.le(0))
+            warnings.warn(f"Some of ranks are zero: {num_zero}")
+            ranks = ranks[ranks.gt(0)]
+
+        # rank should not exceed the number of options
+        if torch.sum(ranks.ge(num_opts + 1)) > 0:
+            num_ge = torch.sum(ranks.ge(num_opts + 1))
+            warnings.warn(f"Some of ranks > 100: {num_ge}")
+            ranks = ranks[ranks.le(num_opts + 1)]
+        return ranks
+
     def get_ranks(self, sample_list, model_output, *args, **kwargs):
         output = model_output["scores"]
         expected = sample_list["targets"]
@@ -499,7 +516,7 @@ class RecallAt1(RecallAtK):
             torch.FloatTensor: Recall@1
 
         """
-        return self.calculate(sample_list, model_output, k=1)
+        return super().calculate(sample_list, model_output, k=1)
 
 
 @registry.register_metric("r@5")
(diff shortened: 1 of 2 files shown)


mmf/modules/metrics.py
```diff
@@ -459,6 +459,22 @@ def get_gt_ranks(self, ranks, ans_ind):
             gt_ranks[i] = int(ranks[i, ans_ind[i].long()])
         return gt_ranks
 
+    def process_ranks(self, ranks):
+        num_opts = 100
+
+        # none of the values should be 0, there is gt in options
+        if torch.sum(ranks.le(0)) > 0:
+            num_zero = torch.sum(ranks.le(0))
+            print(f"Warning: some of ranks are zero: {num_zero}")
```
ytsheng, line 468: let's try to use the warning module, and not have print statements.

ytsheng (commented):
Thanks for adding tests! I have some small nits, but otherwise it looks good!

mmf/modules/metrics.py
brettallenyo: That's a great idea, just changed it :) thanks!

vedanuj (approved):
This looks great! Can you import to Phabricator?

mmf/modules/metrics.py
```diff
@@ -459,6 +460,22 @@ def get_gt_ranks(self, ranks, ans_ind):
             gt_ranks[i] = int(ranks[i, ans_ind[i].long()])
         return gt_ranks
 
+    def process_ranks(self, ranks):
+        num_opts = 100
+
+        # none of the values should be 0, there is gt in options
+        if torch.sum(ranks.le(0)) > 0:
+            num_zero = torch.sum(ranks.le(0))
+            warnings.warn(f"Warning: some of ranks are zero: {num_zero}")
```
apsdehal, line 469: Don't need to mention "Warning:", that will be added automatically by warning module

mmf/modules/metrics.py
```diff
@@ -459,6 +460,22 @@ def get_gt_ranks(self, ranks, ans_ind):
...
+            ranks = ranks[ranks.gt(0)]
+
+        # rank should not exceed the number of options
+        if torch.sum(ranks.ge(num_opts + 1)) > 0:
+            num_ge = torch.sum(ranks.ge(num_opts + 1))
+            warnings.warn(f"Warning: some of ranks > 100: {num_ge}")
```
apsdehal, line 475: Same here.

apsdehal (approved):
Looks good! Thanks for fixing this.

Issue #626, opened 2020-10-14
Training the movie+mcan model using the default config cannot reproduce the reported test-dev score on VQA.


yangapku:
## ❓ Questions and Help
Hi, I have trained a movie+mcan model on my own server using the default config yaml (batchsize=32, update_frequency=2, step=236000). Now I have many checkpoints which are saved every 1k steps. May I ask which one among them should be picked for inference and achieve better performance compared with others according to your experience? Will the movie_mcan_final.pth just be okay? Thank you very much!

vedanuj:
For the movie+mcan submission to VQA Challenge we picked the last checkpoint. You can also use the `movie_mcan_final.pth`.

yangapku:
@vedanuj  Hi, I used the generated movie_mcan_final.pth checkpoint for inference. However, the test-dev score is much lower than expected (test-dev "yes/no": 86.96, "number": 56.23, "other": 62.52, "overall": 71.87). Can you help to figure out where the problem is? Thank you!

yangapku:
The training config: [link](https://yangan.oss-cn-beijing.aliyuncs.com/mmf_checkpoint/rerun_bs32_uf2_fixbug_seed16304007-16310113/config.yaml).

yangapku:
I also tried the last checkpoint (236000 steps), the performance is slightly better (test-dev yes/no: 87.16, number: 56.78, other: 63.01, overall: 72.25) but still not satisfying. Are there some implementation details I missed? Thank you very much!

vedanuj:
You are using batch_size=32 and update_frequency=2 and step=236000. You are essentially doing half the number of updates to the model since the max_updates=236000 which is kept constant as used with batch_size=64 in our original runs. You will need to double the number of steps or double your batch_size.

yangapku:
Thank you very much! Since the 16GB memory of my GPUs cannot hold batch_size=64, I think I should try to double the max_updates.

yangapku:
@vedanuj Hi, I noticed that in the implementation codes, it seems that the `num_updates` increases 1 when the batch passed `update_frequency` times. So does it mean the model essentially performed `max_updates` times of updates with "real batch size" `batch_size * update_frequency` rather than half of that?

vedanuj:
Sorry my bad! You are right. The number of epochs should be same irrespective of `update_frequency`. I will have to retrain and check why it is not giving the same result.

yangapku:
[...]

PR #747, opened 2021-01-23
[fix] fix gradient accumulation when update_frequency != 1

ytsheng:
* fix gradient accumulation when update_frequency is not equal to 1
  * The issue is caused by "start_update" being called at an irregular time so the gradients are actually not being accumulated in between update_frequency. When optimizer's `zero_grad` is called, grad is manually set to zero and therefore the gradients are not being accumulated. This fix made it so that `zero_grad` is called in the correct interval so that grad are being accumulated in between update_frequency.   
* fix `combined_report` not taking into account the new losses causing metrics calculation to be incorrect.
* add test (including logging)
  * add test to make sure there is a final update even if the num batches from the current update < update frequency.
* address issue: https://github.com/facebookresearch/mmf/issues/626

### DIFF
--- a/mmf/common/report.py
+++ b/mmf/common/report.py
@@ -74,7 +74,7 @@ class Report(OrderedDict):
     def fields(self):
         return list(self.keys())
 
-    def accumulate_tensor_fields(self, report, field_list):
+    def accumulate_tensor_fields_and_loss(self, report, field_list):
         for key in field_list:
             if key not in self.keys():
                 warnings.warn(
@@ -84,3 +84,16 @@ class Report(OrderedDict):
                 continue
             if isinstance(self[key], torch.Tensor):
                 self[key] = torch.cat((self[key], report[key]), dim=0)
+
+        self._accumulate_loss(report)
+
+    def _accumulate_loss(self, report):
+        for key, value in report.losses.items():
+            if key not in self.losses.keys():
+                warnings.warn(
+                    f"{key} not found in report. Loss calculation "
+                    + "might not work as expected."
+                )
+                continue
+            if isinstance(self.losses[key], torch.Tensor):
+                self.losses[key] += value
--- a/mmf/trainers/core/training_loop.py
+++ b/mmf/trainers/core/training_loop.py
@@ -2,6 +2,7 @@
 
 import gc
 import logging
+import math
 import warnings
 from abc import ABC
 from typing import Any, Dict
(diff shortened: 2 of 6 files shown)


vedanuj:
Thanks for the cleaner implementation. Where was the bug actually?

ytsheng:
The issue is caused by "start_update" being called at an irregular time so the gradients are actually not being accumulated in between update_frequency. When optimizer's `zero_grad` is called, grad is manually set to zero and therefore the gradients are not being accumulated. This fix made it so that `zero_grad` is called in the correct interval so that grad are being accumulated in between update_frequency.

vedanuj:
Thanks for the fix. Good catch! Is it possible to start a run with update frequency =2 for the movie mcan model? Just to make sure we get the desired result with this fix and can close the issue as well.

mmf/trainers/core/training_loop.py
```diff
@@ -56,37 +56,19 @@ def run_training_epoch(self) -> None:
...
+            should_start_update = True
             for idx, batch in enumerate(self.train_loader):
-
-                if (idx + 1) % self.training_config.update_frequency == 0:
+                if should_start_update:
                     combined_report = None
-                    num_batches_for_this_update = min(
-                        self.training_config.update_frequency, num_remaining_batches
-                    )
-
```
vedanuj, line 82: Is the case for remaining batches < self.training_config.update_frequency handled?

ytsheng: @vedanuj, very good point - I think here the default behavior should be not doing the update if the number of iterations in the current update has not reached update_frequency yet -- bc users of this framework probably will not expect the last batch to have a high variance in their loss values.

What I mean is this: If the update_frequency is 100, and the number of batches we currently have done is 210, this means that we have already done 2 updates (2 * 100). And since 10 < 100, we will not be doing another update. Because the loss accumulated from the 10 iterations could be of high variance, which is less desirable to training especially having it be the last update. 

Let me know if you think otherwise. :)

ytsheng:
Sample run in progress...

mmf/trainers/core/training_loop.py
ytsheng: I looked into pytorch lightning's [implementation](https://github.com/PyTorchLightning/pytorch-lightning/blob/e1c152b13c60f064a312f97ad0e326319bfda047/pytorch_lightning/trainer/training_loop.py#L853). It seems like they do perform a final update. Will update this PR.

ytsheng:
@vedanuj, updated! :D Thanks for taking a look. Added a test too!

ytsheng:
@vedanuj, at iteration 232000/236000, I am getting validation accuracy of 0.9078.

mmf/common/report.py
```diff
@@ -84,3 +84,14 @@ def accumulate_tensor_fields(self, report, field_list):
                 continue
             if isinstance(self[key], torch.Tensor):
                 self[key] = torch.cat((self[key], report[key]), dim=0)
+
+    def accumulate_loss(self, report):
+        for key, value in report.losses.items():
+            if key not in self.losses.keys():
+                warnings.warn(
+                    f"{key} not found in report. Metrics calculation "
+                    + "might not work as expected."
```
vedanuj, line 93: Modify the warning message.

mmf/models/base_model.py
```diff
@@ -169,7 +169,6 @@ def __call__(self, sample_list, *args, **kwargs):
         sample_list = to_device(sample_list, model_device)
 
         model_output = super().__call__(sample_list, *args, **kwargs)
-
```
apsdehal, line 172: Needed?

mmf/trainers/core/training_loop.py
```diff
@@ -95,16 +95,22 @@ def run_training_epoch(self) -> None:
                     combined_report.accumulate_tensor_fields(
                         report, self.metrics.required_params
                     )
+                    # loss needs to be accumulated correctly for metric calculation
+                    combined_report.accumulate_loss(report)
```
apsdehal, line 99: It would make sense to create a new `accumulate` function which does both as both are needed always. This will make sure both are never decoupled again.

tests/test_utils.py
```diff
@@ -125,6 +125,19 @@ def forward(self, prepared_batch):
         return model_output
 
 
+class SimpleModelLogits(torch.nn.Module):
```
apsdehal, line 128: Inherit the SimpleModel and add stuff on top of it? (MSELoss doesn't even need to be assigned to module, you can directly use functional: https://pytorch.org/docs/stable/nn.functional.html#mse-loss)

apsdehal (approved):
Thanks for fixing this. Some minor comments and should be good to land.

vedanuj (approved):
Looks good. Please address Aman's comments before landing.

PR #749, opened 2021-01-25
[enhancement] replace usage of PathManager to use ioPath

brettallenyo:
Replace code in file_io to use the new ioPath Path Manager instead of fvcore.

Test Plan:
- use existing `test_file_io.py` test suite

### DIFF
--- a/mmf/utils/file_io.py
+++ b/mmf/utils/file_io.py
@@ -4,12 +4,10 @@ import os
 import shutil
 from typing import List, Optional
 
+from iopath.common.file_io import PathManager as ioPathManagerClass
 
-try:
-    from fvcore.common.file_io import PathManager as FVCorePathManager
-except ImportError:
-    FVCorePathManager = None
 
+ioPathManager = ioPathManagerClass()
 
 try:
     # [FB only] register internal file IO handlers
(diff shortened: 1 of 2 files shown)


ytsheng:
Looks good to me - but the tests are failing.

vedanuj:
> Looks good to me - but the tests are failing.

Tests seem to be failing for unrelated reason. One localized narratives json link is broken : 

```
E       AssertionError: The url https://storage.googleapis.com/localized-narratives/annotations/flickr30k_test_localized_narratives.jsonl is broken. If this is not your own url, please open up an issue on GitHub
```

mmf/utils/file_io.py
```diff
@@ -6,10 +6,11 @@
 
 
 try:
-    from fvcore.common.file_io import PathManager as FVCorePathManager
-except ImportError:
-    FVCorePathManager = None
+    from iopath.common.file_io import PathManager as ioPathManagerClass
```
vedanuj, line 9: We will need to add `iopath` to the `requirements.txt` file as it is open sourced now. Then this try-except block can be removed. You can fix the iopath to the version you are testing with now. Also let's import to Phabricator to see if all the internal tests pass.

vedanuj (changes requested):

mmf/utils/file_io.py
brettallenyo: I updated `requirements.txt` to reflect the new package and updated `file_io.py` to get rid of the try-except block. Then I merged the two commits into one. Now it is in Phabricator but I am not sure why tests would be failing.

brettallenyo:
I could be reading the log files wrong but it seems for a lot of the tests this is the error: 
```
ImportError: DLL load failed while importing win32file: The specified module could not be found.
```

PR #751, opened 2021-01-26
[enhancement] remove external download tests

ytsheng:
* split the test zoo urls into external and internal
* disable external since localized narratives link broke (now it has been fixed though)
* enable internal links

### DIFF
--- a/tests/configs/test_zoo_urls_external.py
+++ b/tests/configs/test_zoo_urls_external.py
@@ -1,0 +1,17 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import unittest
+
+from tests.configs.test_zoo import TestConfigs
+from tests.test_utils import skip_if_macos, skip_if_no_network
+
+
+class TestConfigsExternal(TestConfigs):
+    @skip_if_no_network
+    @skip_if_macos
+    @unittest.skip("External downloads tests manually turned off")
+    def test_zoos(self):
+        self._test_all_zoos(callback=self._check_download, url_type="external")
+
+    def test_sha256sums(self):
+        self._test_all_zoos(callback=self._check_sha256sum, url_type="external")
--- a/tests/configs/test_zoo.py
+++ b/tests/configs/test_zoo.py
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+
 import re
 import time
 import typing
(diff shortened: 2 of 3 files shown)


vedanuj (approved):
Thanks for fixing this. Minor nit : we can consolidate the test classes into one file, since they are all related.

tests/configs/test_zoo_urls_external.py
```diff
@@ -0,0 +1,17 @@
...
+import unittest
+
+from tests.configs.test_zoo import TestConfigs
+from tests.test_utils import skip_if_macos, skip_if_no_network
+
+
+class TestConfigsExternal(TestConfigs):
+    @skip_if_no_network
+    @skip_if_macos
+    @unittest.skip("External downloads tests manually turned off")
```
apsdehal, line 12: I would disagree here that we should disable these tests. Point of these tests is to capture if something is broken including a dataset download which means user would be facing the problem instead of us. I am against disabling these and instead would want the problem to be fixed at the core level by either fixing the download urls or disabling the dataset itself so that user doesn't face this problem. I am okay with CI tests and badge saying failing until it gives us good signal that something is broken.

apsdehal (changes requested):

tests/configs/test_zoo_urls_external.py
ytsheng: This came up yesterday and @vedanuj mentioned that we will do this check in our nightly builds but disabled to be run in our unit test. In this case, what merited the discussion was that Localized Narratives links were broken - and we can't fix it. So it was failing pretty much all PRs' CI tests.

apsdehal: Two things:
1) this is unrelated to nightly builds if this fails it means it is failing for all of the versions regardless of where it started failing so it would be good to know about this as soon as possible.
2) if we still choose to disable it in favor of nightly tests, let's not land this PR until we have nightly tests.

These tests give useful signal that can prevent a broken user experience and I would recommend not disabling these just for the sake of fixing ci.

Here is an alternative solution. Move these tests to a different work flow that is independent of other CI tests that would fix the problem of signal we are getting from broken tests. We can adjust our settings to allow to land a PR without these tests passing.

apsdehal: About the discussion, these tests were the only reason we got to know about broken LN tests, without these user would become our QA tester. If these tests fail, instead we should stop supporting the dataset due to which tests are failing until the links are fixed. This will not fix the root problem instead would just make us oblivious to the existence of the problem.

ytsheng:
Closing this for now.

PR #752, opened 2021-01-26
[feat] Changes to support training on TPUs

ultrons:
Creating this new PR since pytorch-xla branch is in use for an ongoing project.
I have incorporated most of the changes understood from the last comments on https://github.com/facebookresearch/mmf/pull/693
Happy to follow up on the open comments here.

### DIFF
--- a/mmf/common/sample.py
+++ b/mmf/common/sample.py
@@ -396,7 +396,14 @@ def to_device(
     if isinstance(device, str):
         device = torch.device(device)
 
-    if not torch.cuda.is_available():
+    # default value of device_type is cuda
+    # since other device types such as xla can be passed
+    # falling back to cpu should only happen when device_type
+    # is set to cuda but cuda is not available.
+    if device.type == "cuda" and not torch.cuda.is_available():
+        warnings.warn(
+            "Selected device is cuda, but it is NOT available!!!" "Falling back on cpu."
+        )
         device = torch.device("cpu")
     # to_device is specifically for SampleList
     # if user is passing something custom built
--- a/mmf/common/sample.py
+++ b/mmf/common/sample.py
@@ -397,8 +397,8 @@ def to_device(
         device = torch.device(device)
 
     # default value of device_type is cuda
-    # since other device types such as xla can be passed
-    # falling back to cpu should only happen when device_type
+    # Other device types such as xla can also be passed.
+    # Fall back to cpu only happens when device_type
     # is set to cuda but cuda is not available.
     if device.type == "cuda" and not torch.cuda.is_available():
         warnings.warn(
--- a/mmf/datasets/multi_dataset_loader.py
+++ b/mmf/datasets/multi_dataset_loader.py
@@ -8,7 +8,12 @@ import warnings
 
 import numpy as np
 from mmf.utils.build import build_dataloader_and_sampler, build_dataset
-from mmf.utils.distributed import broadcast_scalar, is_dist_initialized, is_master
+from mmf.utils.distributed import (
+    broadcast_scalar,
+    is_dist_initialized,
+    is_master,
+    is_xla,
+)
 from mmf.utils.general import get_batch_size, get_current_device
 from omegaconf import OmegaConf
 
@@ -189,9 +194,19 @@ class MultiDatasetLoader:
     def __len__(self):
         # Since, this is iterator, we need to return total length == number of batches
         batch_size = get_batch_size()
-        # This assumes drop_last=False for all loaders. See also
-        # build_dataloader_and_sampler().
-        return (self._total_length + batch_size - 1) // batch_size
+        # Changed the length to accomadate drop_last == True
+        # drop_last is required if the batch is split into multiple cores
+        # some of the cores may not have enough examples.
+        if is_xla():
+            logging.info(
+                "drop_last is set to True to avoid uneven dimension shapes "
+                "across cores."
+            )
+            return (self._total_length) // batch_size
+        else:
+            # This assumes drop_last=False for all loaders. See also
+            # build_dataloader_and_sampler().
+            return (self._total_length + batch_size - 1) // batch_size
 
     def __iter__(self):
         if self._num_datasets == 1:
--- a/mmf/trainers/core/evaluation_loop.py
+++ b/mmf/trainers/core/evaluation_loop.py
@@ -44,6 +44,7 @@ class TrainerEvaluationLoopMixin(ABC):
 
             combined_report.metrics = self.metrics(combined_report, combined_report)
             self.update_meter(combined_report, meter, eval_mode=True)
+            logger.info("Validation Done")
 
             # enable train mode again
             self.model.train()
@@ -61,7 +62,7 @@ class TrainerEvaluationLoopMixin(ABC):
 
                 for batch in tqdm.tqdm(dataloader):
                     prepared_batch = reporter.prepare_batch(batch)
-                    prepared_batch = to_device(prepared_batch, torch.device("cuda"))
+                    prepared_batch = to_device(prepared_batch, self.device)
                     with torch.cuda.amp.autocast(enabled=self.training_config.fp16):
                         model_output = self.model(prepared_batch)
                     report = Report(prepared_batch, model_output)
--- a/mmf/trainers/core/reporting.py
+++ b/mmf/trainers/core/reporting.py
@@ -50,10 +50,6 @@ class TrainerReportingMixin(ABC):
             if torch.is_tensor(val):
                 if val.dim() == 1:
                     val = val.mean()
-
-            if hasattr(val, "item"):
-                val = val.item()
-
             meter_update_dict.update({key: val})
             total_val += val
 
--- a/mmf/trainers/core/training_loop.py
+++ b/mmf/trainers/core/training_loop.py
@@ -11,7 +11,6 @@ from mmf.common.report import Report
 from mmf.common.sample import to_device
 from mmf.utils.distributed import is_xla
 from mmf.utils.general import clip_gradients, get_max_updates
-from mmf.utils.distributed import is_xla
 from torch import Tensor
 
 
--- a/mmf/trainers/core/training_loop.py
+++ b/mmf/trainers/core/training_loop.py
@@ -11,6 +11,7 @@ from mmf.common.report import Report
 from mmf.common.sample import to_device
 from mmf.utils.distributed import is_xla
 from mmf.utils.general import clip_gradients, get_max_updates
+from mmf.utils.distributed import is_xla
 from torch import Tensor
 
 
--- a/mmf/trainers/core/training_loop.py
+++ b/mmf/trainers/core/training_loop.py
@@ -9,6 +9,7 @@ import torch
 from mmf.common.registry import registry
 from mmf.common.report import Report
 from mmf.common.sample import to_device
+from mmf.utils.distributed import is_xla
 from mmf.utils.general import clip_gradients, get_max_updates
 from torch import Tensor
 
@@ -170,7 +171,7 @@ class TrainerTrainingLoopMixin(ABC):
     def _forward(self, batch: Tensor) -> Dict[str, Any]:
         prepared_batch = self.dataset_loader.prepare_batch(batch)
         # Move the sample list to device if it isn't as of now.
-        prepared_batch = to_device(prepared_batch, torch.device("cuda"))
+        prepared_batch = to_device(prepared_batch, self.device)
         self.profile("Batch prepare time")
         # Arguments should be a dict at this point
 
@@ -199,6 +200,11 @@ class TrainerTrainingLoopMixin(ABC):
                 self.config,
                 scale=self.scaler.get_scale(),
             )
+        if is_xla():
+            import torch_xla.core.xla_model as xm
+
+            # Assumes no model parallel
+            xm.reduce_gradients(self.optimizer)
 
         self.scaler.step(self.optimizer)
         self.scaler.update()
--- a/mmf/utils/build.py
+++ b/mmf/utils/build.py
@@ -11,10 +11,16 @@ from mmf.common import typings as mmf_typings
 from mmf.common.registry import registry
 from mmf.datasets.processors.processors import Processor
 from mmf.utils.configuration import Configuration
-from mmf.utils.distributed import is_dist_initialized
+from mmf.utils.distributed import is_dist_initialized, is_master, is_xla, synchronize
 from mmf.utils.general import get_optimizer_parameters
 from omegaconf import DictConfig, OmegaConf
 
+
+try:
+    import torch_xla.core.xla_model as xm  # noqa
+    import torch_xla.distributed.parallel_loader as pl  # noqa
+except ImportError:
+    xm = None
 
 ProcessorType = Type[Processor]
 ProcessorDict = Dict[str, ProcessorType]
@@ -77,7 +83,22 @@ def build_model(
 
     if hasattr(model, "build"):
         model.load_requirements()
-        model.build()
+        """ Model build involves checkpoint loading
+        If the checkpoint is not available the underlying
+        methods try to download it.
+        Let master build the model (download the checkpoints) while
+        other ranks wait for the sync message
+        Once the master has downloaded the checkpoint and built the
+        model it sends the sync message, completing the synchronization
+        now other cores can proceed to build the model
+        using already downloaded checkpoint.
+        """
+        if is_master():
+            model.build()
+            synchronize()
+        else:
+            synchronize()
+            model.build()
         model.init_losses()
 
     return model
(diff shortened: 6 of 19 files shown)


ytsheng:
Hi @ultrons! Thanks for your first PR! 

Do you mind taking a look at our contributing [guidelines](https://github.com/facebookresearch/mmf/blob/master/.github/CONTRIBUTING.md) and fixing the test failures?

ultrons:
Thanks @ytsheng, I looked at the ci logs. All the fails are complaining 'no module torchvision'. I am not familiar with circleci test setup. Can you provide some hints on where to look for fixing this?

vedanuj:
Hi @ultrons, can you rebase with latest master? I think the CI failures are due to very old base branch.

vedanuj:
Also let's close the PR #693 and rename the title of this.

mmf/datasets/multi_dataset_loader.py
```diff
@@ -186,9 +186,19 @@ def _infer_dataset_probabilities(self):
     def __len__(self):
         # Since, this is iterator, we need to return total length == number of batches
         batch_size = get_batch_size()
+        # Changed the length to accomadate drop_last == True
+        # drop_last is required if the batch is split into multiple cores
+        # some of the cores may not have enough examples.
+        if is_xla():
+            logging.info(
+                "drop_last is set to True to avoid uneven dimension shapes"
```
apsdehal, line 194:
```suggestion
                "drop_last is set to True to avoid uneven dimension shapes "
```

mmf/trainers/core/evaluation_loop.py
```diff
@@ -61,7 +62,7 @@ def prediction_loop(self, dataset_type: str) -> None:
 
                 for batch in tqdm.tqdm(dataloader):
                     prepared_batch = reporter.prepare_batch(batch)
-                    prepared_batch = to_device(prepared_batch, torch.device("cuda"))
+                    prepared_batch = to_device(prepared_batch, self.device)
```
apsdehal, line 65: Good catch!

mmf/trainers/core/reporting.py
```diff
@@ -51,8 +51,8 @@ def update_dict(self, meter_update_dict, values_dict):
                 if val.dim() == 1:
                     val = val.mean()
 
-            if hasattr(val, "item"):
-                val = val.item()
+            #if hasattr(val, "item"):
+            #    val = val.item()
```
apsdehal, line 55: Let's remove this?

mmf/trainers/core/training_loop.py
```diff
@@ -188,6 +187,7 @@ def _start_update(self):
 
     def _backward(self, loss: Tensor) -> None:
         self.scaler.scale(loss).backward()
+
```
apsdehal, line 190: Needed?

mmf/utils/build.py
```diff
@@ -10,10 +10,15 @@
 from mmf.common.registry import registry
 from mmf.datasets.processors.processors import Processor
 from mmf.utils.configuration import Configuration
-from mmf.utils.distributed import is_dist_initialized
+from mmf.utils.distributed import is_dist_initialized, is_xla, is_master, synchronize
 from mmf.utils.general import get_optimizer_parameters
 from omegaconf import DictConfig, OmegaConf
 
+try:
+    import torch_xla.core.xla_model as xm
```
apsdehal, line 18: Does `torch_xla` come with any stringent dependencies that would cause MMF installation to fail in certain scenarios? If not maybe, we can just add it to requirements and assume xla would be there.

Though, let's take a look at this later, I will also need to check what's the status of xla in fb internal.

mmf/utils/build.py
```diff
@@ -76,7 +81,12 @@ def build_model(
 
     if hasattr(model, "build"):
         model.load_requirements()
-        model.build()
+        if is_master():
+            model.build()
+            synchronize()
+        else:
+            synchronize()
+            model.build()
```
apsdehal, line 89: Maybe add a comment here to make it more clear about what's happening.

mmf/utils/checkpoint.py
```diff
@@ -379,11 +384,15 @@ def _get_vcs_fields(self):
             "git/diff": self.git_repo.git.diff("--no-prefix"),
         }
 
+    def save_func(self, *args):
+        return xm.save(*args) if is_xla() else torch.save(*args)
+
     def save(self, update, iteration=None, update_best=False):
         # Only save in main process
-        if not is_master():
+        if not is_master() and not is_xla():
```
apsdehal, line 392: Does this mean all processes in xla will save checkpoints? I am not familiar with xla internal semantics and may be missing something here, but is this the expected behavior?

mmf/utils/checkpoint.py
```diff
@@ -379,11 +384,15 @@ def _get_vcs_fields(self):
...
             return
 
+        logger.info("Checkpoint save operation started!")
```
apsdehal, line 395: I believe this is printed somewhere else as well. No need to repeat.

mmf/utils/checkpoint.py
```diff
@@ -438,22 +447,27 @@ def save(self, update, iteration=None, update_best=False):
...
 
         if update_best:
+            logger.info("Saving best checkpoint")
             with PathManager.open(best_ckpt_filepath, "wb") as f:
-                torch.save(ckpt, f)
+                self.save_func(ckpt, f)
 
         # Save current always
+
+        logger.info("Saving Current checkpoint")
```
apsdehal, line 459:
```suggestion
        logger.info("Saving current checkpoint")
```

mmf/utils/checkpoint.py
```diff
@@ -438,22 +447,27 @@ def save(self, update, iteration=None, update_best=False):
...
-            torch.save(ckpt, f)
+            self.save_func(ckpt, f)
 
         # Remove old checkpoints if max_to_keep is set
         if self.max_to_keep > 0:
             if len(self.saved_iterations) == self.max_to_keep:
                 self.remove(self.saved_iterations.pop(0))
             self.saved_iterations.append(update)
 
+        logger.info("Checkpoint save operation finished!")
```
apsdehal, line 469: Similar to above comment, it is fine to keep it in debug mode maybe?

mmf/utils/distributed.py
```diff
@@ -169,6 +196,11 @@ def byte_tensor_to_object(byte_tensor, max_size=4094):
 def infer_init_method(config):
     if config.distributed.init_method is not None:
         return
+
+    if getattr(config.training, 'device', 'cuda') == 'xla':
+        global _USE_XLA
```
apsdehal, line 201: I would suggest that we use MMF's registry to track globals.

mmf/utils/metsumm.py
```diff
@@ -0,0 +1,8 @@
+def metsumm(stepno=''):
```
apsdehal, line 1: This should go into general utils.

mmf_cli/run.py
```diff
@@ -96,6 +97,7 @@ def run(opts: typing.Optional[typing.List[str]] = None, predict: bool = False):
     if config.distributed.init_method is None:
         infer_init_method(config)
 
+
```
apsdehal, line 100: Needed?

mmf_cli/run.py
```diff
@@ -32,6 +32,7 @@ def main(configuration, init_distributed=False, predict=False):
     if init_distributed:
         distributed_init(config)
 
+
```
apsdehal, line 35: Needed?

requirements.txt
```diff
@@ -1,5 +1,3 @@
-torch==1.6.0
-torchvision==0.7.0
```
apsdehal, line 2: Can we bring these back?

apsdehal (changes requested):
Thanks! This is starting to look great!

Some comments overall:
- Please install pre-commit and lint all of the files.
- Please rebase against the master.
- Edit the title to reflect what the PR adds and close PR #693. 
- I have requested some changes around: global variables, function locations etc.

mmf/datasets/multi_dataset_loader.py
ultrons: incorporated in the next commit. Thanks.

mmf/trainers/core/evaluation_loop.py
ultrons: Credit goes to Francesc :)

mmf/trainers/core/reporting.py
ultrons: Removed in the next commit, thanks.

mmf/trainers/core/training_loop.py
ultrons: Removed.

mmf/utils/build.py
ultrons: This is pattern used in fairseq, There are some discussion to make this seemless. torch-xla does not yet have a pypy package, therefore adding to requirement.txt can cause friction for some users.

mmf/utils/checkpoint.py
ultrons: Added comment explaining the rationale. for xla we use xm.save which saves only the checkpoint only for master and does the synchronization as well. if we run it only only for master, it will result in synchronization lock because other cores wont run the same code.

mmf/utils/checkpoint.py
ultrons: I think forgot to remove it :(, will push this change with the next feedback.

mmf/utils/checkpoint.py
ultrons: Change as suggested.

mmf/utils/checkpoint.py
ultrons: Haven't looked into the debug mode yet. Will look into tomorrow and make changes accrodingly.

mmf/utils/metsumm.py
ultrons: A teammate suggested that this function is not mmf specific and likely should be part of main xla library. So removed it for the time being. If we see the need we can add it as you suggested.

mmf_cli/run.py
ultrons: Removed.

mmf_cli/run.py
ultrons: Removed.

requirements.txt
ultrons: Brought back the the missing lines.

mmf/common/sample.py
```diff
@@ -396,7 +396,14 @@ def to_device(
     if isinstance(device, str):
         device = torch.device(device)
 
-    if not torch.cuda.is_available():
+    # default valude of device_type is cuda
```
apsdehal, line 399:
```suggestion
    # default value of device_type is cuda
```

mmf_cli/run.py
```diff
@@ -108,15 +108,26 @@ def run(opts: typing.Optional[typing.List[str]] = None, predict: bool = False):
...
-            nprocs=config.distributed.world_size,
-        )
+        if is_xla():
+            import torch_xla.distributed.xla_multiprocessing as xmp
+
+            torch.multiprocessing.set_sharing_strategy("file_system")
+            xmp.spawn(
+                fn=distributed_main,
+                args=(configuration, predict),
+                nprocs=8,  # use all 8 TPU cores
```
apsdehal, line 118: noob question: Does this number always stays fixed or should be configurable? Is there a situation where it needs to be configured?

apsdehal (commented):
This is almost there! Thank you for the hardwork. One minor thing around global variable then we should be ready to land.

mmf_cli/run.py
ultrons: Hey Aman, this branch should be activated only when distributed world size > 1. For TPUs you can either use 1 core or 8 cores or a larger pod slice (32, 128, 256, 512, 1024, 2048). For single core xmp.spawn is not used and main is run directly. For larger size, per host (VM) we will use all 8 cores and hence this number is fixed at 8.

mmf/common/sample.py
ultrons: Corrected, thanks.

mmf/utils/distributed.py
ultrons: I changed to use the registry. It resulted into circular reference:
``` File "/home/sivaibhav/mmf/mmf_cli/run.py", line 9, in <module>
  File "/home/sivaibhav/mmf/mmf/__init__.py", line 5, in <module>
    from mmf import utils, common, modules, datasets, models
  File "/home/sivaibhav/mmf/mmf/common/__init__.py", line 3, in <module>
    from .registry import registry
  File "/home/sivaibhav/mmf/mmf/common/registry.py", line 27, in <module>
    from mmf.utils.env import setup_imports
  File "/home/sivaibhav/mmf/mmf/utils/env.py", line 13, in <module>
    from mmf.utils.general import get_absolute_path
  File "/home/sivaibhav/mmf/mmf/utils/general.py", line 10, in <module>
    from mmf.utils.distributed import get_rank, get_world_size
  File "/home/sivaibhav/mmf/mmf/utils/distributed.py", line 13, in <module>
ImportError: cannot import name 'registry'
```
I attempted to break the circular import by pushing get_absolute_path into the function where it was used. Please review.

mmf/utils/distributed.py
```diff
@@ -169,6 +200,12 @@ def byte_tensor_to_object(byte_tensor, max_size=4094):
 def infer_init_method(config):
     if config.distributed.init_method is not None:
         return
+
+    if getattr(config.training, "device", "cuda") == "xla":
+        registry.register("is_xla", True)
+    else:
+        registry.register("is_xla", False)
```
apsdehal, line 207:
```suggestion
    registry.register("is_xla", config.training.get("device", "cuda") == "xla")
```

apsdehal (approved):
LGTM! One minor comment and then we can land.

mmf/utils/env.py
```diff
@@ -39,6 +38,8 @@ def import_user_module(user_dir: str):
     Args:
         user_dir (str): directory which has to be imported
     """
+    from mmf.utils.general import get_absolute_path  # noqa
```
apsdehal, line 41: This works.

ultrons: Thanks Aman.

mmf/utils/distributed.py
ultrons: I agree. Much better. Thanks for suggestion. Just committed your proposed change.

mmf/utils/build.py
```diff
@@ -165,6 +186,17 @@ def build_dataloader_and_sampler(
     if not isinstance(dataset_instance, torch.utils.data.IterableDataset):
         other_args = _add_extra_args_for_dataloader(dataset_instance, other_args)
 
+    if is_xla():
+        dataset_type = dataset_instance.dataset_type
```
apsdehal, line 190: `dataset_type` is not used.

ultrons: I think it's a left over from some intermediate changes. I will remove it.

mmf/common/sample.py
```diff
@@ -396,7 +396,14 @@ def to_device(
     if isinstance(device, str):
         device = torch.device(device)
 
-    if not torch.cuda.is_available():
+    # default value of device_type is cuda
+    # since other device types such as xla can be passed
+    # falling back to cpu should only happen when device_type
+    # is set to cuda but cuda is not available.
```
vedanuj, line 402: This comment is a bit difficult to parse. Should there be a full stop at the end of `since other device types such as xla can be passed`?

mmf/datasets/multi_dataset_loader.py
```diff
@@ -189,9 +194,19 @@ def _infer_dataset_probabilities(self):
...
+        # Changed the length to accomadate drop_last == True
+        # drop_last is required if the batch is split into multiple cores
+        # some of the cores may not have enough examples.
+        if is_xla():
+            logging.info(
+                "drop_last is set to True to avoid uneven dimension shapes "
+                "across cores."
+            )
+            return (self._total_length) // batch_size
+        else:
```
vedanuj, line 206: Are the if and else statements changing the drop_last behavior? Should we make them consistent?

mmf/utils/build.py
```diff
@@ -165,6 +186,16 @@ def build_dataloader_and_sampler(
     if not isinstance(dataset_instance, torch.utils.data.IterableDataset):
         other_args = _add_extra_args_for_dataloader(dataset_instance, other_args)
 
+    if is_xla():
+        shuffle = True
```
vedanuj, line 190: This needs to be always set to True?

vedanuj (approved):
Looks good to me overall. Left a few comments.

mmf/common/sample.py
ultrons: Hey Vedanuj, thank you for reviewing this PR. I agree that comment could have been written better. I have rephrased it a little bit. Pushing the update shortly, please let me know if it reads better, else I am happy to go by your suggestion.

mmf/datasets/multi_dataset_loader.py
ultrons: With XLA we recommend drop_last to be true mainly to avoid uneven last batch, else the user can also incorporate dummy batches to ensure that it is not needed. In this cases I opted with droping the last batch but did not want to touch the default (GPU path). So yes case it does create a difference wrt xla and non-xla branch.

mmf/utils/build.py
ultrons: It's good callout, thank you. I think we should use others["shuffle'] instead of creating a local variable and forcing to True always. I just made this change, pushing shortly.

ultrons: done.

PR #755, opened 2021-01-29
[feat] PL mvp1: validation + logging

ytsheng:
- [x] MVP 1. Evaluation and Logging - Eval accuracy parity between mmf_trainer and lightning_trainer
  -  [X] Validation at the right frequency and at the end (also calculate the right metrics)
  -  [X] Do logging (same logging format with slight difference is ok)
- [ ] Tests - tests only done for pytorch lightning integration
  - [ ] test logging values are the same as `mmf_trainer`'s logging values
  - [ ] test eval values are the same as `mmf_trainer`'s eval values
  - [ ] test tensorboard logging the same as `mmf_trainer`'s

### DIFF
--- a/mmf/trainers/lightning_core/loop_callback.py
+++ b/mmf/trainers/lightning_core/loop_callback.py
@@ -2,23 +2,40 @@
 
 import logging
 
+import torch
+from mmf.common.meter import Meter
 from mmf.common.registry import registry
+from mmf.common.report import Report
+from mmf.trainers.core.reporting import TrainerReportingMixin
+from mmf.utils.logger import calculate_time_left, summarize_report
+from mmf.utils.timer import Timer
 from pytorch_lightning.callbacks.base import Callback
 
 
 logger = logging.getLogger(__name__)
 
 
-class LightningLoopCallback(Callback):
+class LightningLoopCallback(Callback, TrainerReportingMixin):
     def __init__(self, lightning_trainer):
         super().__init__()
         self.lightning_trainer = lightning_trainer
+        self.trainer_config = lightning_trainer.trainer_config
+        self.training_config = lightning_trainer.training
+
+        # for logging
+        self.total_timer = Timer()
+        self.snapshot_iterations = len(lightning_trainer.data_module.val_loader)
+        self.snapshot_iterations //= self.training_config.batch_size
 
     def on_init_start(self, trainer):
         pass
 
     def on_train_start(self, trainer, pl_module):
         registry.register("current_epoch", trainer.current_epoch)
+        self.train_combined_report = None
+
+        self.train_timer = Timer()
+        self.snapshot_timer = Timer()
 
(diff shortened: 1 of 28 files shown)


mmf/trainers/lightning_core/loop_callback.py
```diff
@@ -2,35 +2,196 @@
...
     def __init__(self, lightning_trainer):
         super().__init__()
         self.lightning_trainer = lightning_trainer
+        self.trainer_config = lightning_trainer.trainer_config
+        self.training_config = lightning_trainer.training
+
+        # for logging
+        self.total_timer = Timer()
+        self.snapshot_iterations = len(lightning_trainer.data_module.val_loader)
+        self.snapshot_iterations //= self.training_config.batch_size
```
ananthsub, line 28: could you set these attributes afterwards in on_train_start or on_test_start? there you have access to the trainer, so you don't need to pass it as an input to the constructor here. otherwise you need to pass the trainer into the callback constructor and the callback into the trainer's constructor. not sure how that's working

mmf/trainers/lightning_core/loop_callback.py
```diff
@@ -2,35 +2,196 @@
...
+        logger.info("Stepping into final validation check")
+        # Only do when run_type has train as it shouldn't happen on validation and
+        # inference runs. Inference will take care of this anyways. Also, don't run
+        # if current iteration is divisble by snapshot interval as it will just
+        # be a repeat
+        if (
+            "train" in self.lightning_trainer.run_type
+            and trainer.global_step % self.trainer_config.val_check_interval != 0
+        ):
+            self._start_eval(trainer)
```
ananthsub, line 71: i don't quite follow why evaluation needs to be started like this. why isn't the validation loop sufficient?

it seems like this callback is mixing both the logging of event times (could the profiler be used instead?) along with the evaluation hook. imo these should be separated

mmf/trainers/lightning_core/loop_callback.py
```diff
@@ -2,35 +2,196 @@
...
+    def _start_eval(self, trainer):
+        trainer.test(
+            model=trainer.model,
+            test_dataloaders=self.lightning_trainer.data_module.val_loader,
+        )
+
+    def _save_checkpoint(self, trainer):
+        logger.info("Checkpoint time. Saving a checkpoint.")
+        return
+        # TODO: sash Needs implementation - next mvp
```
ananthsub, line 148: why doesn't the existing model checkpoint callback work?

ananthsub (commented):
I don't quite understand the role of the loop callback as it seems to be:
- handling checkpointing
- Logging timings
- Starting evaluations

Why not add multiple callbacks and let them run in parallel?

ytsheng:
Closing in favor of: https://github.com/facebookresearch/mmf/pull/768

PR #757, opened 2021-01-31
[fix] Fix accumulate tensor fields in evaluation

vedanuj:
Fixes evaluation loop after changes in #747


ytsheng (approved):

ytsheng:
Thanks for the fix!

PR #759, opened 2021-02-01
[feat] Sharded DDP

vedanuj:
Adding ShardedDDP implementation

- Activates ShardedDDP when Optimizer State Sharding(OSS) is enabled.
- Works with both FP16 and FP32
- Requires fairscale to be installed

Test plan:

MMF Transformer VQA accuracy 44k iters (512 batch size): 

FP16 + OSS ShardedDDP + ShadedGradScaler : 65.87, mem: 17573.0 MB

FP16 : 65.56, mem : 19032.0 MB

OSS ShardedDDP: 65.90 , mem : 21073.0 MB

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,10 +22,12 @@ install_dep: &install_dep
       command: |
         source activate mmf
         pip install --upgrade setuptools
+        pip install --upgrade torch torchvision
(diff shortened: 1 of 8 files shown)


ytsheng:
I would love to see the test plan. :) Trying to do the pytorch lightning equivalent recently.

PR #764, opened 2021-02-03
[feat] Add feature extraction script

brettallenyo:
Add a feature extraction script that uses
Huggingface's Fast RCNN structure and code
to discover features in an image.

`extract_features_frcnn.py` is my script, the other
files were taken from Huggingface, other than
super minor changes by me, including rotating
the RGB/BGR values in the preprocessing script.

The script takes a number of flags that
alter it's behavior like exclude files,
input files, output files, batch size,
etc.

Run `python3 mmf/tools/scripts/features/frcnn/extract_features_frcnn.py --model_file model.bin --config_file config.yaml --image_dir ./example_images --output_folder ./output_features`

### DIFF
--- a/tools/scripts/features/frcnn/extract_features_frcnn.py
+++ b/tools/scripts/features/frcnn/extract_features_frcnn.py
@@ -1,0 +1,315 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+"""
+  Run with for example:
+  python3 mmf/tools/scripts/features/frcnn/extract_features_frcnn.py \
+  --model_file model.bin --config_file config.yaml --image_dir \
+  ./example_images --output_folder ./output_features
+"""
+
+import argparse
+import copy
+import logging
+import os
+
+import numpy as np
+import torch
+
+from tools.scripts.features.extraction_utils import chunks, get_image_files
+from tools.scripts.features.frcnn.frcnn_utils import Config
+from tools.scripts.features.frcnn.modeling_frcnn import GeneralizedRCNN
+from tools.scripts.features.frcnn.processing_image import Preprocess
+
+
+class FeatureExtractor:
+
+    MODEL_URL = {"LXMERT": "unc-nlp/frcnn-vg-finetuned"}
+
+    def __init__(self):
+        self.args = self.get_parser().parse_args()
+        self.frcnn, self.frcnn_cfg = self._build_detection_model()
+
+    def get_parser(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            "--model_name",
+            default="LXMERT",
+            type=str,
+            help="Model to use for detection",
+        )
+        parser.add_argument(
+            "--model_file",
+            default=None,
+            type=str,
+            help="Huggingface model file. This overrides the model_name param.",
+        )
+        parser.add_argument(
+            "--config_file", default=None, type=str, help="Huggingface config file"
+        )
+        parser.add_argument(
+            "--start_index", default=0, type=int, help="Index to start from "
+        )
+        parser.add_argument("--end_index", default=None, type=int, help="")
+        parser.add_argument("--batch_size", type=int, default=1, help="Batch size")
+        parser.add_argument(
+            "--num_features",
+            type=int,
+            default=100,
+            help="Number of features to extract.",
+        )
+        parser.add_argument(
+            "--output_folder", type=str, default="./output", help="Output folder"
+        )
+        parser.add_argument("--image_dir", type=str, help="Image directory or file")
+        # TODO add functionality for this flag
+        parser.add_argument(
+            "--feature_name",
+            type=str,
+            help="The name of the feature to extract",
+            default="fc6",
+        )
+        parser.add_argument(
+            "--exclude_list",
+            type=str,
+            help="List of images to be excluded from feature conversion. "
+            + "Each image on a new line",
+            default="./list",
+        )
+        parser.add_argument(
+            "--confidence_threshold",
+            type=float,
+            default=0,
+            help="Threshold of detection confidence above which boxes will be selected",
+        )
+        # TODO finish background flag
+        parser.add_argument(
+            "--background",
+            action="store_true",
+            help="The model will output predictions for the background class when set",
+        )
+        parser.add_argument(
+            "--padding",
+            type=str,
+            default=None,
+            help="You can set your padding, i.e. 'max_detections'",
+        )
+        parser.add_argument(
+            "--visualize",
+            type=bool,
+            default=False,
+            help="Add this flag to save the extra file used for visualization",
+        )
+        parser.add_argument(
+            "--partition",
+            type=int,
+            default=None,
+            help="Add this flag to save the extra file used for visualization",
+        )
+        parser.add_argument(
+            "--max_partition",
+            type=int,
+            default=None,
+            help="Add this flag to save the extra file used for visualization",
+        )
+        return parser
+
+    def _build_detection_model(self):
+        if self.args.config_file:
+            frcnn_cfg = Config.from_pretrained(self.args.config_file)
+        else:
+            frcnn_cfg = Config.from_pretrained(
+                self.MODEL_URL.get(self.args.model_name, self.args.model_name)
+            )
+        if self.args.model_file:
+            frcnn = GeneralizedRCNN.from_pretrained(
+                self.args.model_file, config=frcnn_cfg
+            )
+        else:
+            frcnn = GeneralizedRCNN.from_pretrained(
+                self.MODEL_URL.get(self.args.model_name, self.args.model_name),
+                config=frcnn_cfg,
+            )
+
+        return frcnn, frcnn_cfg
+
+    def get_frcnn_features(self, image_paths):
+        image_preprocess = Preprocess(self.frcnn_cfg)
+
+        images, sizes, scales_yx = image_preprocess(image_paths)
+
+        output_dict = self.frcnn(
+            images,
+            sizes,
+            scales_yx=scales_yx,
+            padding=None,
+            max_detections=self.frcnn_cfg.max_detections,
+            return_tensors="pt",
+        )
+
+        return output_dict
+
+    def _save_feature(self, file_name, full_features, feat_list, info_list):
+        file_base_name = os.path.basename(file_name)
+        file_base_name = file_base_name.split(".")[0]
+        full_feature_base_name = file_base_name + "_full.npy"
+        feat_list_base_name = file_base_name + ".npy"
+        info_list_base_name = file_base_name + "_info.npy"
+        if self.args.visualize:
+            np.save(
+                os.path.join(self.args.output_folder, full_feature_base_name),
+                full_features,
+            )
+        np.save(
+            os.path.join(self.args.output_folder, feat_list_base_name),
+            feat_list.cpu().numpy(),
+        )
+        np.save(os.path.join(self.args.output_folder, info_list_base_name), info_list)
+
+    def _process_features(self, features, index):
+        feature_keys = [
+            "obj_ids",
+            "obj_probs",
+            "attr_ids",
+            "attr_probs",
+            "boxes",
+            "sizes",
+            "preds_per_image",
+            "roi_features",
+            "normalized_boxes",
+        ]
+        single_features = dict()
+
+        for key in feature_keys:
+            single_features[key] = features[key][index]
+
+        confidence = self.args.confidence_threshold
+        idx = 0
+        while idx < single_features["obj_ids"].size()[0]:
+            removed = False
+            if (
+                single_features["obj_probs"][idx] < confidence
+                or single_features["attr_probs"][idx] < confidence
+            ):
+                single_features["obj_ids"] = torch.cat(
+                    [
+                        single_features["obj_ids"][0:idx],
+                        single_features["obj_ids"][idx + 1 :],
+                    ]
+                )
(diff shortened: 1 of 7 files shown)


tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,577 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
vedanuj, line 1: Please remove the unnecessary functions from this. We should only keep the bare minimum that is required to run the script. Also the config and model files can be saved in our FB cluster and downloaded used from there similar to how the extract_features_vmb file is doing.

tools/scripts/features/frcnn/processing_image.py
```diff
@@ -0,0 +1,167 @@
...
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+from PIL import Image
+
+from utils import img_tensorize
+
+
+class ResizeShortestEdge:
```
vedanuj, line 28: Please check if this is similar to `mmf/datasets/processors/image_processors.py` `ResizeShortest` method. If so let's use that.

tools/scripts/features/frcnn/processing_image.py
```diff
@@ -0,0 +1,167 @@
...
+                img = img.permute(2, 0, 1).unsqueeze(0)  # 3, 0, 1)  # hw(c) -> nchw
+                img = F.interpolate(
+                    img, (newh, neww), mode=self.interp_method, align_corners=False
+                ).squeeze(0)
+            img_augs.append(img)
+
+        return img_augs
+
+
+class Preprocess:
```
vedanuj, line 75: What this is doing can potentially be done by `TorchvisionTransforms` class in `mmf/datasets/processors/image_processors.py`. If so let us use that directly instead of adding new functions. One good way to verify if the two implementations can give same results is to write a test and check the same input gives same results for both. Also you can add this test to our test suite.

tools/scripts/features/frcnn/modeling_frcnn.py
```diff
@@ -0,0 +1,2072 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
vedanuj, line 1: For this task this looks good to add it here. 

We need to think how to add this functionality as an encoder (similar to encoders here : [link](https://github.com/facebookresearch/mmf/blob/master/mmf/modules/encoders.py#L194) ), so that this frcnn encoder can take an input image and return the feature directly to a multimodal model. That will enable us to run those models using this frcnn encoder end to end, which will be part of the next milestone 2 task. Something to keep in mind.

tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -0,0 +1,256 @@
...
+import torch
+
+from modeling_frcnn import GeneralizedRCNN
+from processing_image import Preprocess
+from utils import Config
+
+
+class FeatureExtractor:
+
+    MODEL_URL = {"LXMERT": "unc-nlp/frcnn-vg-finetuned"}
```
vedanuj, line 18: You can download the model and use it directly, that will allow us to get rid of lot of the model download logic.

tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -0,0 +1,256 @@
...
+            "--start_index", default=0, type=int, help="Index to start from "
+        )
+        parser.add_argument("--end_index", default=None, type=int, help="")
+        parser.add_argument("--batch_size", type=int, default=2, help="Batch size")
+        parser.add_argument(
+            "--output_folder", type=str, default="./output", help="Output folder"
+        )
+        parser.add_argument("--image_dir", type=str, help="Image directory or file")
+        parser.add_argument(
+            "--feature_index",
```
vedanuj, line 51: Is this the feature type "fc6" or "fc7" etc?

tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -0,0 +1,256 @@
...
+            "obj_ids",
+            "obj_probs",
+            "attr_ids",
+            "attr_probs",
+            "boxes",
+            "sizes",
+            "preds_per_image",
+            "roi_features",
+            "normalized_boxes",
+        ]
```
vedanuj, line 129: We also need to make sure here if the name of the fields saved for these features are same as the name of the fields saved by `extract_features_vmb.py` features. That will ensure we can load the features and train the models without any hassle.

vedanuj (changes requested):
This is superb! I have added some high level comments. Let's address them first before we take a much closer look.

tools/scripts/features/frcnn/extract_features_frcnn.py
brettallenyo: I have it downloaded and I pass it in as a command line argument. However the code seems to find the path to the model after it is first downloaded, should I still leave this for first time users to use?

Basically if you have nothing downloaded and you first run this it will download the model and store it somewhere, the next time you run it it just uses that cached model.

tools/scripts/features/frcnn/extract_features_frcnn.py
brettallenyo: This is what I replaced feature type with, I guess I did not entirely understand what feature type represented so I replaced it with what I assumed it meant.

This takes the index of the feature we are trying to extract (71 is for bike given our current model, etc.)

vedanuj: Feature type is not equivalent to object type. Features can be of two types depending on from which layer of the FRCNN model they are extracted. If it is the last layer then it is fc7 and if it is from the previous layer it is fc6.

tools/scripts/features/frcnn/extract_features_frcnn.py
vedanuj: Yeah we should check where in the code it is trying to donwload for first time users. Basically what I am saying is if we anyway package this model and config and provide it to users they don't need the download functionality. But if the code is calling it from somewhere we need to check from where and remove that call.

tools/scripts/features/frcnn/utils.py
brettallenyo: Removed unnecessary functions, will upload model and config to AWS later

tools/scripts/features/frcnn/processing_image.py
brettallenyo: See https://github.com/facebookresearch/mmf/pull/764#discussion_r571711583

tools/scripts/features/frcnn/processing_image.py
brettallenyo: Update on this: we tried it and it produced a slightly different output for some reason, will need to further explore the potential of replacing this code

vedanuj (commented):
We are close to getting this ready to land. Please rebase this with latest master.

Additionally you can add the command to run this feature extraction as comments on top of the script. If any models are required to be downloaded please mention that in the steps.

tools/scripts/features/frcnn/processing_image.py
ytsheng: This was resolved? Curious if we know why the output is different?

tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -0,0 +1,296 @@
...
+            "cls_prob": probs,
+            "image_width": width,
+            "image_height": height,
+        }
+
+        return single_features, feat_list, info_list
+
+    def _chunks(self, array, chunk_size):
+        for i in range(0, len(array), chunk_size):
+            yield array[i : i + chunk_size], i
```
ytsheng, line 236: this loop accessing in the array like this makes me feel like it is going to error (index out of bounds). :)

tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -0,0 +1,296 @@
...
+
+        if os.path.isfile(image_dir):
+            features = self.get_frcnn_features([image_dir])
+            self._save_feature(image_dir, features[0])
+        else:
+            files = glob.glob(os.path.join(image_dir, "*.png"))
+            files.extend(glob.glob(os.path.join(image_dir, "*.jpg")))
+            files.extend(glob.glob(os.path.join(image_dir, "*.jpeg")))
+
+            files = {f: 1 for f in files}
```
ytsheng, line 249: You can make files a set here. `set(list_name)`

tools/scripts/features/frcnn/extract_features_frcnn.py
brettallenyo: What is actually nice about this is because we are using an array range, you won't get an index out of bounds error. I could be misunderstanding what you're saying though, in which case I can change this and also the `extract_features_vmb.py` version.

For example, if our array is [1, 2, 3] and I run `print(array[4])` it will print out of bounds, but if I run `print[2:4]` it will run successfully.

Other option is I could create a test file for this method? Just a thought

tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -0,0 +1,351 @@
...
+import logging
+import math
+import os
+
+import numpy as np
+import torch
+
+from modeling_frcnn import GeneralizedRCNN
+from processing_image import Preprocess
+from utils import Config
```
apsdehal, line 22: These probably should be full paths, `tools.scripts.features.frcnn.modeling_frcnn`, others similarly. Relative paths cause issues in pip packages.

tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,344 @@
...
+from pathlib import Path
+from urllib.parse import urlparse
+from zipfile import ZipFile, is_zipfile
+
+import cv2
+import numpy as np
+import requests
+from filelock import FileLock
+from PIL import Image
+from yaml import Loader, dump, load
```
apsdehal, line 35: As discussed, we should use OmegaConf here.

tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,344 @@
...
+    from torch.hub import _get_torch_home
+
+    torch_cache_home = _get_torch_home()
+except ImportError:
+    torch_cache_home = os.path.expanduser(
+        os.getenv(
+            "TORCH_HOME", os.path.join(os.getenv("XDG_CACHE_HOME", "~/.cache"), "torch")
+        )
+    )
+
```
apsdehal, line 56: We do have this somewhere in MMF already.

tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,344 @@
...
+CONFIG = os.path.join(PATH, "config.yaml")
+ATTRIBUTES = os.path.join(PATH, "attributes.txt")
+OBJECTS = os.path.join(PATH, "objects.txt")
+PYTORCH_PRETRAINED_BERT_CACHE = os.getenv(
+    "PYTORCH_PRETRAINED_BERT_CACHE", default_cache_path
+)
+PYTORCH_TRANSFORMERS_CACHE = os.getenv(
+    "PYTORCH_TRANSFORMERS_CACHE", PYTORCH_PRETRAINED_BERT_CACHE
+)
+TRANSFORMERS_CACHE = os.getenv("TRANSFORMERS_CACHE", PYTORCH_TRANSFORMERS_CACHE)
```
apsdehal, line 69: If we can use MMF's cache here, it would be great!

tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,344 @@
...
+    assert isinstance(im, str)
+    if os.path.isfile(im):
+        img = cv2.imread(im)
+    else:
+        img = get_image_from_url(im)
+        assert img is not None, f"could not connect to: {im}"
+    img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+    if input_format == "RGB":
+        img = img[:, :, ::-1]
+    return img
```
apsdehal, line 344: Duplicates? Some other functions are present in other PR as well.

tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,344 @@
...
+                tar_file.extractall(output_path_extracted)
+                tar_file.close()
+            else:
+                raise OSError(
+                    f"Archive format of {output_path} could not be identified"
+                )
+
+        return output_path_extracted
+
+    return output_path
```
apsdehal, line 305: We should try to use MMF's download functionalities here. But we can tackle this in Milestone 2 as well. Do mark it.

tools/scripts/features/frcnn/extract_features_frcnn.py
brettallenyo: I just tried it and it suddenly caused a `ModuleNotFoundError: No module named 'pytorch_lightning'` error, any obvious reason for this? I will investigate

apsdehal: `pip install pytorch_lightning`. I think the PL diff landed after you started this.

tools/scripts/features/frcnn/utils.py
brettallenyo: Yes I have them in both but they are used in both, my plan was to just change it for whichever I merge second, presumably the visualization PR

vedanuj (approved):
Looks good to me! Thanks for working on this.

PR #766, opened 2021-02-05
[feat] add visualization API and notebook

brettallenyo:
Add a visualization API for the frcnn
features, which takes in image and feature
paths and uses the huggingface visualization
tools to show the image.

Also added a jupyter notebook that will display
the images accordingly if the features and images
are in the correct place.

### DIFF
--- a/mmf/utils/features/visualizing_image.py
+++ b/mmf/utils/features/visualizing_image.py
@@ -1,0 +1,546 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+"""
+ coding=utf-8
+ Copyright 2018, Antonio Mendoza Hao Tan, Mohit Bansal
+ Adapted From Facebook Inc, Detectron2
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.import copy
+ """
+import colorsys
+import io
+import os
+from typing import Union
+
+import matplotlib as mpl
+import matplotlib.colors as mplc
+import matplotlib.figure as mplfigure
+import numpy as np
+import requests
+import torch
+from matplotlib.backends.backend_agg import FigureCanvasAgg
+from mmf.utils.download import get_image_from_url
+from PIL import Image
+
+
+_SMALL_OBJ = 1000
+
+
+def get_data(query: str, delim: str = ","):
+    assert isinstance(query, str)
+    if os.path.isfile(query):
+        with open(query) as f:
+            data = eval(f.read())
+    else:
+        req = requests.get(query)
+        try:
+            data = requests.json()
+        except Exception:
+            data = req.content.decode()
+            assert data is not None, "could not connect"
+            try:
+                data = eval(data)
+            except Exception:
+                data = data.split("\n")
+        req.close()
+    return data
+
+
+def img_tensorize(im: str):
+    assert isinstance(im, str)
+    if os.path.isfile(im):
+        img = np.array(Image.open(im))
+    else:
+        img = get_image_from_url(im)
+        assert img is not None, f"could not connect to: {im}"
+    return img
+
+
+class SingleImageViz:
+    def __init__(
+        self,
+        img: Union[str, np.ndarray],
+        scale: float = 1.2,
+        edgecolor: str = "g",
+        alpha: float = 0.5,
+        linestyle: str = "-",
+        saveas: str = "test_out.jpg",
+        rgb: bool = True,
+        pynb: bool = False,
+        id2obj: list = None,
+        id2attr: list = None,
+        pad: float = 0.7,
+    ):
+        """
+        img: an RGB image of shape (H, W, 3).
+        """
+        if isinstance(img, torch.Tensor):
+            img = img.numpy().astype("np.uint8")
+        if isinstance(img, str):
+            img = img_tensorize(img)
+        assert isinstance(img, np.ndarray)
+
+        width, height = img.shape[1], img.shape[0]
+        fig = mplfigure.Figure(frameon=False)
+        dpi = fig.get_dpi()
+        width_in = (width * scale + 1e-2) / dpi
+        height_in = (height * scale + 1e-2) / dpi
+        fig.set_size_inches(width_in, height_in)
+        ax = fig.add_axes([0.0, 0.0, 1.0, 1.0])
+        ax.axis("off")
+        ax.set_xlim(0.0, width)
+        ax.set_ylim(height)
+
+        np.random.seed(42)  # generate same random
+
+        self.saveas = saveas
+        self.rgb = rgb
+        self.pynb = pynb
+        self.img = img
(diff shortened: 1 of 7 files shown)


brettallenyo:
Screen Shot 2021-02-05 at 12 29 15 PM

The notebook looks like this in case people were wondering

tools/scripts/features/frcnn/ImageVisualizer.ipynb
```diff
@@ -0,0 +1,127 @@
+{
```
apsdehal, line 1: We usually land notebooks in `notebooks` branch and do not land any ipython code in master.

brettallenyo: Ah thats great to know thank you, I will adjust

tools/scripts/features/frcnn/utils.py
```diff
@@ -0,0 +1,577 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
vedanuj, line 1: Is this file required for this PR? Also this has lot of unnecessary code. Let's remove them.

vedanuj (changes requested):
This is great. Excited to see the notebook running. Please restructure the code as per my comments. Also as @apsdehal noted, let's not add the ipython notebook in this PR. We can add it later to the notebooks branch. 

It will be good to keep this PR only about the visualization API that you are adding to the utils. It will be good to write some tests to check working of this API as well.

vedanuj:
> What should I use for the image and feature names to abstract the notebook for all users

The notebook can add the full functionality of extracting the features and displaying the bounding boxes. If the user inputs an image, the notebook will have step by step instructions of extracting the features (you might have to link to the feature extraction apis that you have implemented in the other PR #764), then visualize the boxes that were extracted. 

> Should I include the training and inference command line code so the notebook is a one-stop-shop for everything I have done so far

Training and inference doesn't need to be coupled with this. This notebook can just be about feature extraction and visualization.

> Should I upload the images and features so it works everywhere even without the model training

Not sure I understand the question properly. We will need to extract the features using your script in PR #764 for COCO and Visual Genome datasets. We will store the features locally in FAIR Cluster to train the multimodal models with these extracted features. These models will later be used for end to end inference once we add your frcnn script as an encoder.

mmf/utils/features/visualizing_image.py
```diff
@@ -0,0 +1,517 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.import copy
+ """
+import colorsys
+import io
+import os
+
+import cv2
```
vedanuj, line 21: Let's try to use PIL image library instead of cv2 if possible. Adding cv2 to utils will require additional dependency to mmf core which we should try to avoid if possible.

vedanuj (changes requested):
Some tests are failing. Those are because of adding additional dependencies like cv2. Let's see if we can use PIL reader instead. 

Also rebase this on master and then import this to Phabricator to check if internal tests pass.

mmf/utils/features/visualizing_image.py
```diff
@@ -0,0 +1,517 @@
...
+def img_tensorize(im, input_format="RGB"):
+    assert isinstance(im, str)
+    if os.path.isfile(im):
+        img = cv2.imread(im)
+    else:
+        img = get_image_from_url(im)
+        assert img is not None, f"could not connect to: {im}"
+    img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+    if input_format == "RGB":
+        img = img[:, :, ::-1]
```
ytsheng, line 50: cvtColor is doing exactly `img = img[:, :, ::-1]`, so ifyou do cvtColor(cv2.COLOR_BGR2RGB) twice I believe you will get back the original image. So the logic here seems a bit weird.

ytsheng (commented):
Looks good to me. cv2 is pretty heavy - PIL seems better. :)

mmf/utils/features/visualizing_image.py
brettallenyo: I agree, I think it is some weird cv2 behavior that I just dropped in accordance with Vedanuj's comment about removing that dependency

mmf/utils/features/visualizing_image.py
```diff
@@ -0,0 +1,515 @@
...
+import requests
+import torch
+from matplotlib.backends.backend_agg import FigureCanvasAgg
+from PIL import Image
+
+
+_SMALL_OBJ = 1000
+
+
+def get_image_from_url(url):
```
apsdehal, line 34: I would suggest moving to `mmf/utils/download.py`.

mmf/utils/features/visualizing_image.py
```diff
@@ -0,0 +1,515 @@
...
+    assert isinstance(im, str)
+    if os.path.isfile(im):
+        img = np.array(Image.open(image_path))
+    else:
+        img = get_image_from_url(im)
+        assert img is not None, f"could not connect to: {im}"
+    return img
+
+
+class SingleImageViz:
```
apsdehal, line 50: Do we plan to add typings to this? or some sort of tests?

mmf/utils/features/visualizing_image.py
```diff
@@ -0,0 +1,515 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
apsdehal, line 1: Is there a reason this can't also go into utils/visualize.py?

mmf/utils/visualize.py
```diff
@@ -45,3 +48,27 @@ def visualize_images(
 
     plt.axis("off")
     plt.imshow(grid.permute(1, 2, 0))
+
+
+def visualize_frcnn_features(image_path, features_path, objids, attrids):
+    img = np.array(Image.open(image_path))
```
apsdehal, line 54: Isn't this technically same as `img_tensorize`?

mmf/utils/visualize.py
```diff
@@ -45,3 +48,27 @@ def visualize_images(
...
+
+    output_dict = np.load(features_path, allow_pickle="TRUE").item()
```
apsdehal, line 56:
```suggestion
    output_dict = np.load(features_path, allow_pickle=True).item()
```

mmf/utils/visualize.py
brettallenyo: Actually yes, I am realizing I can actually just remove that line and pass in the image path directly

mmf/utils/features/visualizing_image.py
brettallenyo: I could definitely put it there, I just did not want to overcrowd that file with feature-displaying-specific code, do you think I should move it?

mmf/utils/features/visualizing_image.py
vedanuj: @brettallenyo Please add typings to all the functions in this file.

mmf/utils/visualize.py
```diff
@@ -45,3 +48,27 @@ def visualize_images(
 
     plt.axis("off")
     plt.imshow(grid.permute(1, 2, 0))
+
+
+def visualize_frcnn_features(image_path, features_path, objids, attrids):
```
vedanuj, line 53: Add typings here as well

vedanuj: Might be good to write a test for this method. One way to test is make sure the input and output image arrays are same.

vedanuj (approved):
Looks good to me. Thanks for adding the test. Please rebase the branch before landing.

PR #771, opened 2021-02-11
[feat] PL mvp1: validation + logging

ytsheng:
- [x] MVP 1. Evaluation and Logging - Eval accuracy parity between mmf_trainer and lightning_trainer
  -  [X] Validation at the right frequency and at the end (also calculate the right metrics)
  -  [X] Do logging (same logging format with slight difference is ok)
- [x] Tests - tests only done for pytorch lightning integration
  - [X] test logging values are the same as `mmf_trainer`'s logging values
  - [X] test eval values are the same as `mmf_trainer`'s eval values
  - [x] test tensorboard logging the same as `mmf_trainer`'s
- [ ] Blocked:
  - [ ] One Epoch Evaluation Run after train: [PR4945](https://github.com/PyTorchLightning/pytorch-lightning/pull/4945/files) and [PR4948](https://github.com/PyTorchLightning/pytorch-lightning/pull/4948/files)

### DIFF
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -147,7 +147,7 @@ trainer:
     # https://pytorch-lightning.readthedocs.io/en/latest/trainer.html#trainer-class-api
     type: lightning
     params:
-        gpus: null
+        gpus: 1
         num_nodes: 1
         precision: 32
         deterministic: false
@@ -156,12 +156,14 @@ trainer:
         max_epochs: null
         gradient_clip_val: 0.0
         num_sanity_val_steps: 0
-        automatic_optimization: true # only True is supported for now
         checkpoint_callback: false
         accumulate_grad_batches: 1
         val_check_interval: 1000
         log_every_n_steps: 100
-        limit_val_batches: 5
+        logger: false
+        limit_val_batches: 1.0
+        # progress bar is turned off if want same logging format as `mmf_trainer`
+        progress_bar_refresh_rate: 0
 
 # Configuration for evaluation
 evaluation:
--- a/mmf/models/base_model.py
+++ b/mmf/models/base_model.py
@@ -49,6 +49,7 @@ from typing import List, Optional, Union
 
 import pytorch_lightning as pl
 from mmf.common.registry import registry
+from mmf.common.report import Report
 from mmf.common.sample import SampleList, to_device
 from mmf.modules.losses import LossConfig, Losses
 from mmf.utils.checkpoint import load_pretrained_model
@@ -118,6 +119,12 @@ class BaseModel(pl.LightningModule):
             "Build method not implemented in the child model class."
         )
 
+    def build_meters(self, run_type):
+        from mmf.utils.build import build_meters
+
+        """Function only used in lightning setting"""
+        self.train_meter, self.val_meter, self.test_meter = build_meters(run_type)
+
     def init_losses(self):
         """Initializes loss for the model based ``losses`` key. Automatically called by
         MMF internally after building the model.
@@ -175,7 +182,7 @@ class BaseModel(pl.LightningModule):
             "Forward of the child model class needs to be implemented."
         )
 
-    def training_step(self, batch, batch_idx, *args, **kwargs):
+    def training_step(self, batch: SampleList, batch_idx: int, *args, **kwargs):
         """Member function of PL modules. Used only when PL enabled.
         To be implemented by child class. Takes in a ``SampleList``,
         batch_idx and returns back a dict.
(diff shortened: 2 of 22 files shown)


mmf/trainers/lightning_core/loop_callback.py
```diff
@@ -32,9 +49,37 @@ def on_train_batch_end(
...
+
     def on_train_end(self, trainer: Trainer, pl_module: LightningModule):
-        # TODO: @sash next PR
-        pass
+        # Only do when run_type has train as it shouldn't happen on validation and
+        # inference runs. Inference will take care of this anyways. Also, don't run
+        # if current iteration is divisble by snapshot interval as it will just
+        # be a repeat
+        if (
+            "train" in self.lightning_trainer.run_type
```
apsdehal, line 70: At some places we are using `self.lightning_trainer.x` while at some `self.training_config.x`. This would be confusing for an end user inheriting this to understand where is what.

mmf/trainers/lightning_core/loop_callback.py
```diff
@@ -47,3 +92,134 @@ def on_validation_batch_end(
...
+        ), "mmf lightning_trainer supports 1 optimizer per model for now."
+        optimizer = trainer.optimizers[0]
+        return optimizer
+
+    def _save_checkpoint(self, trainer: Trainer):
+        logger.info("Checkpoint time. Saving a checkpoint.")
+        return
+        # TODO: sash Needs implementation - next mvp
+
+    def _get_current_epoch_for_logging(self, trainer: Trainer):
```
apsdehal, line 167: Shouldn't these be properties?

mmf/utils/general.py
```diff
@@ -333,33 +332,6 @@ def get_sizes_list(dim, chunks):
     return sizes_list
 
 
-def get_max_updates(config_max_updates, config_max_epochs, train_loader, update_freq):
```
apsdehal, line 336: As discussed before can we import these here as well and add `# noqa` to it so that flake8 doesn't complain.

tests/test_utils.py
```diff
@@ -136,7 +136,13 @@ def forward(self, prepared_batch):
         batch = prepared_batch[DATA_ITEM_KEY]
         output = self.linear(batch)
         loss = torch.nn.MSELoss()(-1 * output, batch)
-        return {"losses": {"loss": loss}, "logits": output, "input_batch": input_sample}
+        return {
+            "losses": {"loss": loss},
+            "logits": output,
+            "input_batch": input_sample,
+            "dataset_type": "dummpy_dataset_type",
```
apsdehal, line 143:
```suggestion
            "dataset_type": "dummy_dataset_type",
```

next line as well.

mmf/configs/defaults.yaml
```diff
@@ -147,7 +147,7 @@ trainer:
     # https://pytorch-lightning.readthedocs.io/en/latest/trainer.html#trainer-class-api
     type: lightning
     params:
-        gpus: null
+        gpus: 0
```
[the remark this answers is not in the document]
ytsheng: This makes it so that the default is idx=0 gpu, as opposed to before, on cpu.

mmf/models/base_model.py
```diff
@@ -205,9 +201,28 @@ def validation_step(self, batch, batch_idx, *args, **kwargs):
...
+            Dict
+        """
+        return self._forward_lightning_step(batch, batch_idx)
+
+    def _forward_lightning_step(self, batch, batch_idx):
         batch = self._ensure_sample_list(batch)
         output = self(batch)
-        # TODO: @sash Implementation coming soon! (next PR)
+        loss_dict = output["losses"]
+        output["loss"] = sum(loss.mean() for loss in loss_dict.values())
```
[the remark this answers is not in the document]
ytsheng: `loss` is required, `input_batch` is returned here so we can have access to it in the callback function to compute report and metrics.

mmf/trainers/core/training_loop.py
```diff
@@ -10,7 +10,8 @@
 from mmf.common.report import Report
 from mmf.common.sample import to_device
 from mmf.utils.distributed import is_xla
-from mmf.utils.general import clip_gradients, get_max_updates
+from mmf.utils.flow import extract_loss, get_max_updates
```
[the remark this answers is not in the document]
ytsheng: Since now there are two places who are going to use those functions, I think it is much cleaner to move them to a utils file where both can access. I thought calling it flow makes sense because it could be used for both training and evaluation work flow.

mmf/trainers/lightning_core/loop_callback.py
ytsheng: run_type is a config parameter, so we need to grab the config from the trainer to be able to access this variable. In the init of this class, I added comments to make it clear which is which. I also simplified the logic here by a little bit.

mmf/trainers/lightning_core/loop_callback.py
ytsheng: I do think properties is the best - since it is just + 1 from the original number.

mmf/utils/general.py
ytsheng: ok! Will move them here!

tests/test_utils.py
ytsheng: oh good catch! 💛

ytsheng (commented):
Thanks for review!

PR #774, opened 2021-02-12
[refactor,tests,fix] Clean up and fix a bug in MMFT; add extensive tests

apsdehal:
The diff address a bug that was introduced in text position ids by a
recent change in #760. The fix is to actually make single feature dim
modalities B X 1 X D before passing to backend instead of making
multiple assumption in backend layer which tend to be conflicting with
text input ids. This also adds some comments to better explain stack
version of text.

The PR also refactors MMFT to be more readble and reduce cyclomatic
complexity.

Finally, this PR adds extensive tests to make sure a change like #760
doesn't break MMFT again.

Test Plan:
Adds extensive tests for MMFT

Differential Revision: D26410408

### DIFF
--- a/mmf/models/mmf_transformer.py
+++ b/mmf/models/mmf_transformer.py
@@ -161,66 +161,89 @@ class MMFTransformer(BaseTransformer):
         Check the definition of this dataclass in BaseTransformer.
         """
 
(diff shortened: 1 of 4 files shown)


vedanuj (approved):
Thanks for refactoring this and adding extensive tests. Looks good to me.

PR #798, opened 2021-03-05
[feat] Inference API

brettallenyo:
Added an API to run inference in these
use cases:
- region features -> mmft
- region features -> visual_bert

Based on configs, some are provided here,
users can run inference on various
trained models with something as simple
as an image URL and a text query.

Would love for people to take a look and provide
feedback about my coding style and design choices.

If you want to run it on your own you can pull and run 
```
from mmf.utils.inference import Inference

if __name__ == "__main__":
    inference = Inference(checkpoint_path="/checkpoint/brettallen/visualbert")
    answer = inference.forward(
        "http://i.imgur.com/1IWZX69.jpg",
        {"text": "what type of shoes is the woman in pink wearing"},
        image_format="url",
    )
    print(answer)
```

### DIFF
--- a/mmf/datasets/processors/__init__.py
+++ b/mmf/datasets/processors/__init__.py
@@ -1,7 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 from mmf.datasets.processors.bert_processors import MaskedTokenProcessor
+from mmf.datasets.processors.frcnn_processor import FRCNNPreprocess
 from mmf.datasets.processors.image_processors import TorchvisionTransforms
+from mmf.datasets.processors.prediction_processors import ArgMaxPredictionProcessor
 from mmf.datasets.processors.processors import (
     BaseProcessor,
     BBoxProcessor,
@@ -33,4 +35,6 @@ __all__ = [
     "CaptionProcessor",
     "MaskedTokenProcessor",
     "TorchvisionTransforms",
+    "FRCNNPreprocess",
+    "ArgMaxPredictionProcessor",
 ]
--- a/mmf/datasets/processors/frcnn_processor.py
+++ b/mmf/datasets/processors/frcnn_processor.py
@@ -1,0 +1,186 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+"""
+ coding=utf-8
+ Copyright 2018, Antonio Mendoza Hao Tan, Mohit Bansal
+ Adapted From Facebook Inc, Detectron2
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.import copy
+ """
+import os
+import sys
+from dataclasses import dataclass
+from typing import List
+
+import numpy as np
+import omegaconf
+import torch
+import torch.nn.functional as F
+from mmf.common.registry import registry
+from mmf.datasets.processors.processors import BaseProcessor
+from mmf.utils.download import get_image_from_url
+from PIL import Image
+
+
+class ResizeShortestEdge:
+    def __init__(self, short_edge_length: List[int], max_size: int = sys.maxsize):
+        """
+        Args:
+            short_edge_length (list[min, max])
+            max_size (int): maximum allowed longest edge length.
+        """
+        self.interp_method = "bilinear"
+        self.max_size = max_size
+        self.short_edge_length = short_edge_length
+
+    def __call__(self, imgs: List[torch.Tensor]):
+        img_augs = []
+        for img in imgs:
+            h, w = img.shape[:2]
+            # later: provide list and randomly choose index for resize
+            size = np.random.randint(
+                self.short_edge_length[0], self.short_edge_length[1] + 1
+            )
+            if size == 0:
+                return img
+            scale = size * 1.0 / min(h, w)
+            if h < w:
+                newh, neww = size, scale * w
+            else:
+                newh, neww = scale * h, size
+            if max(newh, neww) > self.max_size:
+                scale = self.max_size * 1.0 / max(newh, neww)
+                newh = newh * scale
+                neww = neww * scale
+            neww = int(neww + 0.5)
+            newh = int(newh + 0.5)
+
+            if img.dtype == np.uint8:
+                pil_image = Image.fromarray(img)
+                pil_image = pil_image.resize((neww, newh), Image.BILINEAR)
+                img = np.asarray(pil_image)
+            else:
+                img = img.permute(2, 0, 1).unsqueeze(0)  # 3, 0, 1)  # hw(c) -> nchw
+                img = F.interpolate(
+                    img, (newh, neww), mode=self.interp_method, align_corners=False
+                ).squeeze(0)
+            img_augs.append(img)
+
+        return img_augs
+
+
+@registry.register_processor("frcnn_preprocess")
+class FRCNNPreprocess(BaseProcessor):
+    @dataclass
+    class FRCNNPreprocessConfig:
+        model: omegaconf.DictConfig = omegaconf.MISSING
+        input: omegaconf.DictConfig = omegaconf.MISSING
+        size_divisibility: int = 0
+        pad_value: float = 0
+
+    def __init__(self, config: FRCNNPreprocessConfig, *args, **kwargs):
+        config_input = config.get("input", None)
+        assert config_input is not None
+        min_size_test = config_input.get("min_size_test", 800)
+        max_size_test = config_input.get("max_size_test", 1333)
+        self.aug = ResizeShortestEdge([min_size_test, min_size_test], max_size_test)
+        self.input_format = config_input.get("format", "BGR")
+        self.size_divisibility = config.get("size_divisibility", 0)
+        self.pad_value = config.get("pad_value", 0)
+        self.max_image_size = max_size_test
+        config_model = config.get("model", None)
+        assert config_model is not None
+        self.device = config_model.get("device", "cpu")
+        config_pixel_std = config_model.get("pixel_std", [1.0, 1.0, 1.0])
+        self.pixel_std = (
+            torch.tensor(config_pixel_std)
+            .to(self.device)
+            .view(len(config_pixel_std), 1, 1)
+        )
+        config_pixel_mean = config_model.get(
+            "pixel_mean", [102.9801, 115.9465, 122.7717]
+        )
+        self.pixel_mean = (
+            torch.tensor(config_pixel_mean)
+            .to(self.device)
+            .view(len(config_pixel_std), 1, 1)
+        )
+        self.normalizer = lambda x: (x - self.pixel_mean) / self.pixel_std
+
+    def pad(self, images: List[torch.Tensor]):
+        max_size = tuple(max(s) for s in zip(*[img.shape for img in images]))
+        image_sizes = [im.shape[-2:] for im in images]
+        images = [
+            F.pad(
+                im,
+                [0, max_size[-1] - size[1], 0, max_size[-2] - size[0]],
+                value=self.pad_value,
+            )
+            for size, im in zip(image_sizes, images)
+        ]
+
+        return torch.stack(images), torch.tensor(image_sizes)
+
+    def __call__(self, images: torch.Tensor, single_image: bool = False):
+        """
+        Takes images of variable sizes, returns preprocessed
+        version based on config sizing, etc.
+        """
+        with torch.no_grad():
+            if not isinstance(images, list):
+                images = [images]
+            if single_image:
+                assert len(images) == 1
+            for i in range(len(images)):
+                if isinstance(images[i], torch.Tensor):
+                    images.insert(i, images.pop(i).to(self.device).float())
+                elif not isinstance(images[i], torch.Tensor):
+                    images.insert(
+                        i,
+                        torch.as_tensor(
+                            img_tensorize(images.pop(i), input_format=self.input_format)
+                        )
+                        .to(self.device)
+                        .float(),
+                    )
+            # resize smallest edge
+            raw_sizes = torch.tensor([im.shape[:2] for im in images])
+            images = self.aug(images)
+
+            # flip rgb to bgr
+            for idx in range(len(images)):
+                images[idx] = torch.flip(images[idx], [0])
+            # transpose images and convert to torch tensors
+            # images = [torch.as_tensor(i.astype("float32"))
+            # .permute(2, 0, 1).to(self.device) for i in images]
+            # now normalize before pad to avoid useless arithmetic
+            images = [self.normalizer(x) for x in images]
+            # now pad them to do the following operations
+            images, sizes = self.pad(images)
+            # Normalize
+
+            if self.size_divisibility > 0:
+                raise NotImplementedError()
+            # pad
+            scales_yx = torch.true_divide(raw_sizes, sizes)
+            if single_image:
+                return images[0], sizes[0], scales_yx[0]
+            else:
+                return images, sizes, scales_yx
+
+
+def img_tensorize(im: str):
+    assert isinstance(im, str)
+    if os.path.isfile(im):
+        img = np.array(Image.open(im))
+    else:
+        img = get_image_from_url(im)
+        assert img is not None, f"could not connect to: {im}"
+    return img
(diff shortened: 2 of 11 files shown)


mmf/datasets/processors/__init__.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 from mmf.datasets.processors.bert_processors import MaskedTokenProcessor
+from mmf.datasets.processors.frcnn_processor import FRCNNPreprocess  # noqa
```
vedanuj, line 4: Remove the noqa comment and format it with isort

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,187 @@
...
+import requests
+import torch
+import torch.nn.functional as F
+from mmf.common.registry import registry
+from mmf.datasets.processors.processors import BaseProcessor
+from PIL import Image
+
+
+class ResizeShortestEdge:
+    def __init__(self, short_edge_length, max_size=sys.maxsize):
```
vedanuj, line 32: Let's add typings to all the functions.

mmf/models/__init__.py
```diff
@@ -8,6 +8,8 @@
 from .top_down_bottom_up import TopDownBottomUp
 from .butd import BUTD
 from .mmbt import MMBT, MMBTForClassification, MMBTForPreTraining
+from .mmf_transformer import MMFTransformer  # noqa
+from .movie_mcan import MoVieMcan  # noqa
```
vedanuj, line 12: Same here. Use isort.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+                tar_file.extractall(output_path_extracted)
+                tar_file.close()
+            else:
+                raise OSError(
+                    f"Archive format of {output_path} could not be identified"
+                )
+
+        return output_path_extracted
+
+    return output_path
```
vedanuj, line 146: If these are not required please remove them.

mmf/modules/encoders.py
```diff
@@ -338,6 +341,47 @@ def forward(self, x):
...
+    def __init__(self, config: Config, *args, **kwargs):
+        super().__init__()
+        self.config = config
+        # pretrained = config.get("pretrained", False)
+        # pretrained_path = config.get("pretrained_path", None)
+
+        if config.pretrained:
+            self.frcnn = GeneralizedRCNN.from_pretrained(
+                config.pretrained_path, config=config
+            )
```
vedanuj, line 361: Let's get rid of the from_pretrained method. You can directly download the checkpoint and then use load_state_dict to load the checkpoint after the model is initialized from config. 

Check how it is done in `Detectron2ResnetImageEncoder` above.

mmf/modules/encoders.py
```diff
@@ -338,6 +341,47 @@ def forward(self, x):
...
+        else:
+            self.frcnn = GeneralizedRCNN(config)
+
+    def forward(
```
vedanuj, line 365: Please add typings to all method arguments.

mmf/utils/inference.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+if __name__ == "__main__":
+    inference = Inference()
+    answer = inference.forward(
+        "http://i.imgur.com/1IWZX69.jpg",
+        {"text": "what type of shoes is the woman in pink wearing"},
+        image_format="url",
+    )
+    print(answer)
```
vedanuj, line 96: I think we should build a new command similar to `mmf_cli` commands for run, predict etc. Basically you can create an inference command which is interactive and prompts user to enter an image url and text.

mmf/utils/inference.py
```diff
@@ -0,0 +1,96 @@
...
+
+        sample_list = SampleList([sample])
+        sample_list = sample_list.to(get_current_device())
+        self.model = self.model.to(get_current_device())
+
+        output = self.model(sample_list)
+
+        answers = output["scores"].argmax(dim=1)
+
+        answer = self.processor["answer_processor"].idx2word(answers[0])
```
vedanuj, line 84: Currently this is not generic and can handle only VQA Q/A type input. What will be a good way to make it generic for all type of tasks(or majority of tasks)? cc @apsdehal

mmf/utils/inference.py
```diff
@@ -0,0 +1,96 @@
...
+from mmf.common.registry import registry  # noqa
+from mmf.common.sample import Sample, SampleList
+from mmf.utils.build import build_encoder, build_model, build_processors
+from mmf.utils.checkpoint import load_pretrained_model
+from mmf.utils.general import get_current_device
+from PIL import Image
+
+
+class Inference:
+    def __init__(self, checkpoint_path=None):
```
vedanuj, line 18: This shouldn't be a script but an API. So what you can do is take the checkpoint_path as an input to init. Let's get rid of the parser.

mmf/utils/inference.py
```diff
@@ -0,0 +1,96 @@
...
+        return processor, feature_extractor, model
+
+    def forward(self, image_path, text, image_format="path"):
+        text_output = self.processor["text_processor"](text)
+        if image_format == "path":
+            img = np.array(Image.open(image_path))
+        elif image_format == "url":
+            img = np.array(Image.open(requests.get(image_path, stream=True).raw))
+        img = torch.as_tensor(img)
+        max_detect = self.config.features.image_feature_encodings.params.max_detections
```
vedanuj, line 51: Move this inside the if condition

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+                self.roi_heads.res5[i].conv2.dilation = (2, 2)
+
+    def forward_for_roi_head(self, features, proposal_boxes):
+        # assert not proposal_boxes[0].requires_grad
+        box_features = self.roi_heads._shared_roi_transform(features, proposal_boxes)
+        feature_pooled = box_features.mean(dim=[2, 3])  # pooled to 1x1
+        return feature_pooled
+
+    @classmethod
+    def from_pretrained(cls, pretrained_model_name_or_path, *model_args, **kwargs):
```
vedanuj, line 369: This method is not required. Check how to load the checkpoint without using `from_pretrained`

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+    @property
+    def out_channels(self):
+        return self.conv1.out_channels
+
+    @property
+    def stride(self):
+        return 4  # = stride 2 conv -> stride 2 max pool
+
+
+class GeneralizedRCNN(nn.Module):
```
vedanuj, line 311: Please add typings to everything. Also use `config` instead of `cfg` to maintain the same convention as other MMF functions.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+        return x.new_empty(new_shape)
+
+    @staticmethod
+    def backward(ctx, grad):
+        shape = ctx.shape
+        return _NewEmptyTensorOp.apply(grad, shape), None
+
+
+# Main Classes
+class Conv2d(torch.nn.Conv2d):
```
vedanuj, line 238: Can these classes not be replaced with what Detectron2 has? Or are these customized for the frcnn script?

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+                    f"Archive format of {output_path} could not be identified"
+                )
+
+        return output_path_extracted
+
+    return output_path
+
+
+# Helper Functions
+def get_norm(norm, out_channels):
```
vedanuj, line 150: Are these functions not available in D2?

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
vedanuj, line 1: For the feature extraction script that you had added earlier you should replace the function calls in that with the functions you have added here. That will help clean up the code inside tools/scripts.

vedanuj (changes requested):
This is a great contribution. Please check my comments and address them.

mmf/datasets/processors/__init__.py
brettallenyo: Whenever I remove noqa it says `mmf/datasets/processors/__init__.py:4:1: F401 'mmf.datasets.processors.frcnn_processor.FRCNNPreprocess' imported but unused` but if I remove the import it doesnt work and I get this error:
Screen Shot 2021-03-19 at 11 32 09 AM

vedanuj: You will need to add it to `__all__ = [... ` below.

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,173 @@
...
+import omegaconf
+import requests
+import torch
+import torch.nn.functional as F
+from mmf.common.registry import registry
+from mmf.datasets.processors.processors import BaseProcessor
+from PIL import Image
+
+
+class ResizeShortestEdge:
```
apsdehal, line 32: We have this already in MMF. Please check if you can reuse: https://github.com/facebookresearch/mmf/blob/8409361b704cd2e90b410a71165a6073caa8f13a/tools/scripts/features/frcnn/processing_image.py#L28

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,173 @@
...
+                    img, (newh, neww), mode=self.interp_method, align_corners=False
+                ).squeeze(0)
+            img_augs.append(img)
+
+        return img_augs
+
+
+@registry.register_processor("frcnn_preprocess")
+class FRCNNPreprocess(BaseProcessor):
+    def __init__(self, cfg: omegaconf.dictconfig.DictConfig, *args, **kwargs):
```
apsdehal, line 81:
```suggestion
    def __init__(self, cfg: omegaconf.DictConfig, *args, **kwargs):
```

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,173 @@
...
+        self.aug = ResizeShortestEdge(
+            [cfg.input.min_size_test, cfg.input.min_size_test], cfg.input.max_size_test
+        )
+        self.input_format = cfg.input.format
```
apsdehal, line 85: You can use `.get` to get config values and then add a default value for user's sanity.

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,173 @@
...
+                img = F.interpolate(
+                    img, (newh, neww), mode=self.interp_method, align_corners=False
+                ).squeeze(0)
+            img_augs.append(img)
+
+        return img_augs
+
+
+@registry.register_processor("frcnn_preprocess")
+class FRCNNPreprocess(BaseProcessor):
```
apsdehal, line 80: Add a dataclass defining input for this.

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,173 @@
...
+                im,
+                [0, max_size[-1] - size[1], 0, max_size[-2] - size[0]],
+                value=self.pad_value,
+            )
+            for size, im in zip(image_sizes, images)
+        ]
+
+        return torch.stack(images), torch.tensor(image_sizes)
+
+    def __call__(self, images: torch.Tensor, single_image: bool = False):
```
apsdehal, line 116: Add docstrings for end user.

mmf/datasets/processors/frcnn_processor.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def img_tensorize(im: str):
+    assert isinstance(im, str)
+    if os.path.isfile(im):
+        img = np.array(Image.open(im))
+    else:
+        img = get_image_from_url(im)
+        assert img is not None, f"could not connect to: {im}"
+    return img
```
apsdehal, line 173: I have seen these two somewhere else as well. Please reuse or move to a image utils and then refactor and reuse.

mmf/modules/encoders.py
apsdehal: Might be useful to keep it as well for future use cases and then use load_state_dict here.

mmf/utils/inference.py
apsdehal: Agreed, script using this API should go to tools.

mmf/utils/inference.py
```diff
@@ -0,0 +1,96 @@
...
+    def get_parser(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            "--checkpoint", default=None, type=str, help="Inference config file"
+        )
+        return parser
+
+    def _build_model(self):
+        model_items = load_pretrained_model(self.checkpoint)
+        self.config = omegaconf.OmegaConf.create(model_items["full_config"])
```
apsdehal, line 35: Please do `from omegaconf import OmegaConf` and then use OmegaConf.

mmf/utils/inference.py
apsdehal: I think we can define something like output_processor which defaults to argmax prediction processor and then can be replaced accordingly.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+
+"""
+ coding=utf-8
+ Copyright 2018, Antonio Mendoza Hao Tan, Mohit Bansal
```
apsdehal, line 5: This can be added to NOTICES as well.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+    from detectron2.modeling import ShapeSpec
+    from detectron2.modeling.backbone.resnet import BottleneckBlock, ResNet
+    from detectron2.modeling.proposal_generator.rpn import RPN, StandardRPNHead
+    from detectron2.modeling.roi_heads.roi_heads import Res5ROIHeads
+    from detectron2.structures.image_list import ImageList
+except ImportError:
+    pass
+
+
+mmf_cache_home = get_mmf_cache_dir()
```
apsdehal, line 50: This is not right. We shouldn't call it outside of functions as `cache_dir` for MMF might not be defined by the time you are calling this. This should probably be converted to a function and used to get the value when needed.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+        v = ckp.pop(k)
+        if isinstance(v, np.ndarray):
+            v = torch.tensor(v)
+        else:
+            assert isinstance(v, torch.tensor), type(v)
+        r[k] = v
+    return r
+
+
+# TODO use mmf's download functionality
```
apsdehal, line 77: Yes, please.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+        norm = {
+            "BN": BatchNorm2d,
+            "GN": lambda channels: nn.GroupNorm(32, channels),
+            "nnSyncBN": nn.SyncBatchNorm,  # keep for debugging
+            "": lambda x: x,
+        }[norm]
+    return norm(out_channels)
+
+
+def build_backbone(cfg):
```
apsdehal, line 163: Doesn't d2 have a build_backbone function? If this is different from that please delineate the difference.

apsdehal: Also add typings to the whole script.

mmf/models/frcnn.py
apsdehal: If they are custom, please document the differences.

mmf/models/frcnn.py
```diff
@@ -0,0 +1,555 @@
...
+        image_list = ImageList(images, image_shapes)
+
+        # generate proposals if none are available
+        if proposals is None:
+            proposal_boxes, _ = self.proposal_generator(image_list, features, gt_boxes)
+        else:
+            assert proposals is not None
+
+        proposal_boxes = [proposal_boxes[0].get_fields()["proposal_boxes"]]
+
```
apsdehal, line 548: The linebreaks are weird here and below.

mmf/models/frcnn.py
brettallenyo: Where should I document the differences? In a comment at the top of the class?

mmf/models/frcnn.py
brettallenyo: Didn't realize it was in D2, just replaced it

mmf/utils/inference.py
brettallenyo: Should that be included in this PR?

mmf/datasets/processors/frcnn_processor.py
brettallenyo: I will replace the usage of that preprocessor to use this

mmf/models/frcnn.py
brettallenyo: We talked about this off github but want it to be here for future ref, the GeneralizedRCNN in tools is different from the one in this commit because it calculates many other values like obj_ids, attr_ids, bounding boxes, etc. So we cannot reference the one here in tools. Hopefully in a future PR we can bring the two together and have just one version.

mmf/utils/inference.py
brettallenyo: Will be done in future PR, going to resolve for now

vedanuj (approved):
Thanks for addressing the comments. Looks good to me.

PR #823, opened 2021-03-23
[feat] add images to localized narratives

ytsheng:


mmf/datasets/builders/localized_narratives/masked_dataset.py
```diff
@@ -35,6 +35,13 @@ def __getitem__(self, idx: int) -> Sample:
                 image_info_0.pop("image_id")
             current_sample.update(features)
 
+        if self._use_images:
```
vedanuj, line 38: Should we have this condition in elif? And else can throw a Unknown type error.

vedanuj (approved):
Thanks. LGTM!

Minor nit. Also some tests are failing.

PR #830, opened 2021-03-25
[feat] Inference CLI

brettallenyo:
Created script to allow users to run
inference using various models, text
inputs, and image inputs.

All that is required is using
mmf_inference and passing some
command line args.

Here is how you use it:
```
mmf_interactive checkpoint_path=/checkpoint/brettallen/mmft/model
```

This will initiate an interactive inference
script that will ask for image URL and text
input to run on. It will keep prompting the user
for more images and text until they say `exit`.

I also added the ability to just say `same` to use
the same image because I figured it was inconvenient
for the user to constantly have to copy over
the image URL every time they want to ask a new
question.

### DIFF
--- a/mmf_cli/interactive.py
+++ b/mmf_cli/interactive.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3 -u
+# Copyright (c) Facebook, Inc. and its affiliates.
+import argparse
+import logging
+import typing
+
+from mmf.utils.configuration import _merge_with_dotlist
+from mmf.utils.flags import flags
+from mmf.utils.inference import Inference
+from mmf.utils.logger import setup_logger
+from omegaconf import OmegaConf
+
+
+def construct_config(opts: typing.List[str]):
+    config = OmegaConf.create({"checkpoint_path": ""})
+    return _merge_with_dotlist(config, opts)
+
+
+def interactive(opts: typing.Optional[typing.List[str]] = None):
+    """Inference runs inference on an image and text provided by the user.
+    You can optionally run inference programmatically by passing an optlist as opts.
+
+    Args:
+        opts (typing.Optional[typing.List[str]], optional): Optlist which can be used.
+            to override opts programmatically. For e.g. if you pass
+            opts = ["checkpoint_path=my/directory"], this will set the checkpoint.
+    """
+    if opts is None:
+        parser = flags.get_parser()
+        args = parser.parse_args()
+    else:
+        args = argparse.Namespace(config_override=None)
+        args.opts = opts
+
+    setup_logger()
+    logger = logging.getLogger("mmf_cli.interactive")
+
+    config = construct_config(args.opts)
+    inference = Inference(checkpoint_path=config.checkpoint_path)
+    logger.info("Enter 'exit' at any point to terminate.")
+    logger.info("Enter an image URL:")
+    image_url = input()
+    logger.info("Got image URL.")
+    logger.info("Enter text:")
+    text = input()
+    logger.info("Got text input.")
+    while text != "exit":
+        logger.info("Running inference on image and text input.")
+        answer = inference.forward(image_url, {"text": text}, image_format="url")
+        logger.info("Model response: " + answer)
+        logger.info(
+            f"Enter another image URL or leave it blank to continue using {image_url}:"
+        )
+        new_image_url = input()
+        if new_image_url != "":
+            image_url = new_image_url
+        if new_image_url == "exit":
+            break
+        logger.info("Enter another text input:")
+        text = input()
+
+
+if __name__ == "__main__":
+    interactive()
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -156,6 +156,7 @@ def fetch_package_data():
                 "mmf_run = mmf_cli.run:run",
                 "mmf_predict = mmf_cli.predict:predict",
                 "mmf_convert_hm = mmf_cli.hm_convert:main",
+                "mmf_inference = mmf_cli.inference:inference",
```
apsdehal, line 159: Tha name mmf_inference is somewhat conflicting with mmf_predict. I think the original idea was to create it in a fashion of interactive mode where the user can keep on inputting text and url and it keeps on predicting.

mmf_cli/inference.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        args = argparse.Namespace(config_override=None)
+        args.opts = opts
+
+    config = construct_config(args.opts)
+    inference = Inference(checkpoint_path=config.checkpoint_path)
+    answer = inference.forward(
+        config.image_url, {"text": config.text}, image_format="url"
+    )
+    print(config.text.capitalize() + "? " + answer)
```
apsdehal, line 46: Use logger.

mmf_cli/inference.py
```diff
@@ -0,0 +1,50 @@
...
+import typing
+
+from mmf.utils.flags import flags
+from mmf.utils.inference import Inference
+from omegaconf import OmegaConf
+
+
+def construct_config(opts: typing.List[str]):
+    config = OmegaConf.create()
+    opt_values = [opt.split("=", maxsplit=1) for opt in opts]
```
apsdehal, line 13: Not necessary but I think you should refactor out the logic in configuration as there is many more cases that you need to handle that we have already done in configuration opts logic parser.

mmf_cli/inference.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    if opts is None:
+        parser = flags.get_parser()
+        args = parser.parse_args()
+    else:
+        args = argparse.Namespace(config_override=None)
+        args.opts = opts
+
+    config = construct_config(args.opts)
+    inference = Inference(checkpoint_path=config.checkpoint_path)
```
apsdehal, line 42: model is suggested as an input in your command, but where is it being used. how does inference figure out based on checkpoint which model it is?

apsdehal (changes requested):
Thanks. This is an amazing contribution.

setup.py
ytsheng: Maybe `mmf-interactive`?

brettallenyo: Changed it to an interactive CLI and changed the name to `mmf_interactive`

mmf_cli/inference.py
brettallenyo: Model is suggested as we are looking to eventually make this work with the zoo, but for now inference figures out which model using the checkpoint's config

mmf_cli/inference.py
brettallenyo: I tried to use the configuration logic but I came across a few problems: configuration can only take fields that are used in the default config, i.e. it cannot take new fields like `checkpoint` or `checkpoint_path` as `checkpoint` is initialized to something else (a dictconfig struct) and `checkpoint_path` is not included in the default config so it just throws an error. The other problem is the configuration also _requires_ certain fields that inference does not require like `datasets`. The way to work around that is to just programmatically include a random dataset but I figured it would just be easier to do this

hivestrung: To deal with the dataset requirement, just modify configuration.py:

```
class Configuration:
    def __init__(self, args=None, default_only=False, load_dataset=True):
...
other_configs = self._build_other_configs(load_dataset=load_dataset)
...
def _build_other_configs(self, load_dataset=True):
        opts_config = self._build_opt_list(self.args.opts)
        user_config = self._build_user_config(opts_config)

        self._opts_config = opts_config
        self._user_config = user_config

        self.import_user_dir()
        model_config = self._build_model_config(opts_config)
        args_overrides = self._build_demjson_config(self.args.config_override)
        other_configs = None
        if load_dataset:
            dataset_config = self._build_dataset_config(opts_config)
            other_configs = OmegaConf.merge(
                model_config, dataset_config, user_config, args_overrides
            )
        else:
            other_configs = OmegaConf.merge(
                model_config, user_config, args_overrides
            )
            

        return other_configs

```


I'm dealing with this issue too as I am using the models for inference, this was the simplest way to deal with it.

mmf_cli/interactive.py
```diff
@@ -0,0 +1,63 @@
...
+    logger.info("Enter text:")
+    text = input()
+    logger.info("Got text input.")
+    while text != "exit":
+        logger.info("Running inference on image and text input.")
+        answer = inference.forward(image_url, {"text": text}, image_format="url")
+        logger.info("Model response: " + answer)
+        logger.info("Enter another image URL or 'same' to use previous one:")
+        new_image_url = input()
+        if new_image_url != "same":
```
apsdehal, line 54: If input is empty then also it means same? Show in the prompt itself what is the value that will be used in `[]`. And if user presses enter meaning empty string then it is same.

apsdehal (approved):
Except minor comment it looks great. Thanks.

PR #837, opened 2021-03-29
[test] reduce mem leak in tests

ytsheng:
* Fix memory leak in pytest
* Test Plan: run `pytest tests`


vedanuj:
LGTM. Please add a description and test plan to this diff before landing.

vedanuj:
You can also add the memory leak profiler which can be useful to catch future regressions.

PR #839, opened 2021-03-30
[fix] Use caption only files for localized narratives

ytsheng:
@vedanuj. This PR uses smaller caption files.

### DIFF
--- a/mmf/configs/zoo/datasets.yaml
+++ b/mmf/configs/zoo/datasets.yaml
@@ -543,6 +543,40 @@ localized_narratives:
         - url: https://storage.googleapis.com/localized-narratives/annotations/open_images_test_localized_narratives.jsonl
           file_name: open_images_test_localized_narratives.jsonl
           compressed: false
+  only_captions:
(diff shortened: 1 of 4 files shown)


mmf/configs/zoo/datasets.yaml
```diff
@@ -543,6 +543,36 @@ localized_narratives:
         - url: https://storage.googleapis.com/localized-narratives/annotations/open_images_test_localized_narratives.jsonl
           file_name: open_images_test_localized_narratives.jsonl
           compressed: false
+        - url: https://storage.googleapis.com/localized-narratives/annotations/open_images_train_v6_captions.jsonl
```
vedanuj, line 546: Should we split this into a separate zoo node like `localized_narratives.only_captions` or something? What do you think?

vedanuj (approved):
Thanks for this! One minor comment.

mmf/configs/zoo/datasets.yaml
ytsheng: sounds good. Will update.

Issue #678, opened 2020-11-10
AttributeError: 'VisualGenomeDataset' object has no attribute '_get_absolute_path'


Anth0nyWu:
## 🐛 Bug

 File "/home/ubuntu/vqa/GMN/mmf/mmf/datasets/builders/visual_genome/dataset.py", line 44, in __init__
    scene_graph_file = self._get_absolute_path(scene_graph_file)
AttributeError: 'VisualGenomeDataset' object has no attribute '_get_absolute_path'

## Command that i run in shell

CUDA_VISIBLE_DEVICES="0" mmf_run config=projects/gmn/configs/visual_genome/defaults.yaml model=gmn datasets=visual_genome run_type=train env.data_dir=/media/ubuntu/MyDisk/data_mmf/vg

## To Reproduce

I ran the command above where gmn is a model class inheriting from Pythia and get this error:

Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/mmf4/bin/mmf_run", line 33, in <module>
    sys.exit(load_entry_point('mmf', 'console_scripts', 'mmf_run')())
  File "/home/ubuntu/vqa/GMN/mmf/mmf_cli/run.py", line 123, in run
    main(configuration, predict=predict)
  File "/home/ubuntu/vqa/GMN/mmf/mmf_cli/run.py", line 52, in main
    trainer.load()
  File "/home/ubuntu/vqa/GMN/mmf/mmf/trainers/mmf_trainer.py", line 41, in load
    super().load()
  File "/home/ubuntu/vqa/GMN/mmf/mmf/trainers/base_trainer.py", line 29, in load
    self.load_datasets()
  File "/home/ubuntu/vqa/GMN/mmf/mmf/trainers/mmf_trainer.py", line 68, in load_datasets
    self.dataset_loader.load_datasets()
  File "/home/ubuntu/vqa/GMN/mmf/mmf/common/dataset_loader.py", line 17, in load_datasets
    self.train_dataset.load(self.config)
  File "/home/ubuntu/vqa/GMN/mmf/mmf/datasets/multi_dataset_loader.py", line 114, in load
    self.build_datasets(config)
  File "/home/ubuntu/vqa/GMN/mmf/mmf/datasets/multi_dataset_loader.py", line 129, in build_datasets
    dataset_instance = build_dataset(dataset, dataset_config, self.dataset_type)
[...]

PR #841, opened 2021-03-30
Included _get_absolute_path

sonuagarwal1008:
AttributeError: 'VisualGenomeDataset' object has no attribute '_get_absolute_path' #678

### DIFF
--- a/mmf/datasets/builders/visual_genome/dataset.py
+++ b/mmf/datasets/builders/visual_genome/dataset.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import copy
 import json
-
+import os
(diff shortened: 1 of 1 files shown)


sonuagarwal1008:
@vedanuj  Initially i replicated the same error in my local, when i executed this command mmf_run config=projects/movie_mcan/configs/vqa2/defaults.yaml model=movie_mcan dataset=visual_genome run_type=train i got the same issue "**function not found**" later added that function and issue got resolved.

sonuagarwal1008:
@vedanuj old function was correct for only one case. That's why was getting the correct output. I am updating that file this time I used the **get_mmf_env** function and joined that with file name. please let me know if I am going in the right direction.

Issue #848, opened 2021-04-02
extract_features_frcnn: _save_feature call not correctly written


hivestrung:
## 🐛 Bug

Function call for _save_features will fail for single image

## To Reproduce
https://github.com/facebookresearch/mmf/blob/9d7696f49932b48f701e56aca4d32ffe279ee28f/tools/scripts/features/frcnn/extract_features_frcnn.py#L266
The function call doesn't have the right arguments...

```
self._save_feature(image_dir, features[0])
# compared to:
self._save_feature(
                            file_names[begin_idx + idx],
                            full_features,
                            feat_list,
                            info_list,
                        )
# function def
def _save_feature(self, file_name, full_features, feat_list, info_list):
        file_base_name = os.path.basename(file_name)
        file_base_name = file_base_name.split(".")[0]
        full_feature_base_name = file_base_name + "_full.npy"
        feat_list_base_name = file_base_name + ".npy"
        info_list_base_name = file_base_name + "_info.npy"
        # before saving, ensure the output dir is created
        os.makedirs(self.args.output_folder, exist_ok=True)
        if self.args.visualize:
            np.save(
                os.path.join(self.args.output_folder, full_feature_base_name),
                full_features,
            )
        np.save(
            os.path.join(self.args.output_folder, feat_list_base_name),
            feat_list.cpu().numpy(),
        )
        np.save(os.path.join(self.args.output_folder, info_list_base_name), info_list)

```
## Who can help

@brettallenyo

brettallenyo:
Thanks so much for the catch, should be fixed with: https://github.com/facebookresearch/mmf/pull/850

hivestrung:
Thanks for the fix, will test it out and close this if it works

hivestrung:
@brettallenyo Seems to be ok, but you might need to modify img_tensorize like so:

Note the .convert("RGB"). I was uploading some of the Hateful Memes image for testing and it turns out they're "RGBA" which required this conversion. But I really don't know what's the best way to go about ensuring that formats are correctly input.

```
def img_tensorize(im: str):
    assert isinstance(im, str)
    if os.path.isfile(im):
        img = np.array(Image.open(im).convert("RGB"))
    else:
        img = get_image_from_url(im)
        assert img is not None, f"could not connect to: {im}"
    return img
```

brettallenyo:
@hivestrung good catch, definitely want it to apply to those use cases, just updated the diff

PR #850, opened 2021-04-02
[fix] FRCNN Feature extraction for single image

brettallenyo:
Fixed single image bug that had incorrect
function arguments for extracting
features on a single image.

Fix: #848

### DIFF
--- a/tools/scripts/features/frcnn/processing_image.py
+++ b/tools/scripts/features/frcnn/processing_image.py
@@ -69,9 +69,7 @@ class Preprocess:
                 elif not isinstance(images[i], torch.Tensor):
                     images.insert(
                         i,
-                        torch.as_tensor(
-                            img_tensorize(images.pop(i), input_format=self.input_format)
-                        )
+                        torch.as_tensor(img_tensorize(images.pop(i)))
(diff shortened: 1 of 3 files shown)


tools/scripts/features/frcnn/processing_image.py
```diff
@@ -69,9 +69,7 @@ def __call__(self, images, single_image=False):
                 elif not isinstance(images[i], torch.Tensor):
                     images.insert(
                         i,
-                        torch.as_tensor(
-                            img_tensorize(images.pop(i), input_format=self.input_format)
```
vedanuj, line 73: Is the `input_format` arg not required?

vedanuj (approved):
LGTM! Please make sure it works for both single as well as multiple images.

tools/scripts/features/frcnn/processing_image.py
brettallenyo: No its not I simplified the function pretty significantly from what huggingface had, they used a cv2 and some other weird functions that I just omitted, I guess I just forgot to delete that param in this use of the function when I first committed.

Issue #836, opened 2021-03-29
tools/scripts/features/frcnn/extract_features_frcnn.py


1144181135:
## ❓ Questions and Help
I try to extract bottom up features in a new dataset. The feature extraction in "mmf/tools/scripts/features/extract_features_vmb.py" is achieved by maskrcnn_benchmark but it shows deprecated. I found that the file "tools/scripts/features/frcnn/extract_features_frcnn.py". Can this file extract bottom-up features? If can, where is the MODEL_URL? If not, how to extract features with detectron2 ?

vedanuj:
Hi .. you can use `tools/scripts/features/frcnn/extract_features_frcnn.py` for extracting features. The script will automatically download the model file and config file from MODEL_URL defined in the script if you don't specify it in the run command.

hivestrung:
The script doesn't actually download the model file and config file from MODEL_URL, but I've been putting together an upgrade that has fixed most of those issues and will upload it in the next week or so.

hivestrung:
@1144181135 Please check out my fork https://github.com/hivestrung/mmf/commit/6332a9803721a9a230913b6e2589fed172ae4778
This commit has code that will most likely solve your problems with using MODEL_URL to download the model

PR #858, opened 2021-04-05
[fix] Download functionality work for frcnn

brettallenyo:
Fixed the downloading functionality
for the FRCNN feature extraction script.

This resolves the TODO of using the
mmf download functionality to download
the configs and models.

Also updated the script to directly use
the huggingface urls

Resolves: #836

### DIFF
--- a/tools/scripts/features/frcnn/extract_features_frcnn.py
+++ b/tools/scripts/features/frcnn/extract_features_frcnn.py
@@ -23,7 +23,15 @@ from tools.scripts.features.frcnn.processing_image import Preprocess
 
 class FeatureExtractor:
 
-    MODEL_URL = {"LXMERT": "unc-nlp/frcnn-vg-finetuned"}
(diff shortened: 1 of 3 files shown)


tools/scripts/features/frcnn/extract_features_frcnn.py
```diff
@@ -23,7 +23,15 @@
 
 class FeatureExtractor:
 
-    MODEL_URL = {"LXMERT": "unc-nlp/frcnn-vg-finetuned"}
+    MODEL_URL = {
+        "LXMERT": "https://s3.amazonaws.com/models.huggingface.co/bert/unc-nlp/"
```
vedanuj, line 27: Model should be named as FRCNN.

vedanuj (changes requested):
Looks good overall.

vedanuj (approved):

Issue #567, opened 2020-09-16
[enhancment] SampleList should check device of tensor being added


apsdehal:
## 🚀 Feature

SampleList in all cases, (update, setattr) should check that the tensor or object being added has same device as sample list itself.

## Motivation

This minor oversight can cause issues later down the lane as .to may not move everything to device to keep checks shallow.
The validation check should happen at the add level itself.

## Pitch

Check device of tensor field if present and ensure newly added tensor has same device.

prachigarg23:
Hi, I would like to work on this. This will be my first attempt at contributing. This feature is still required, right? Just to clarify, this feature will have to be incorporated at all instances of sample_list across the repository right?

apsdehal:
Hi,
Thanks for your interest. This would happen inside the samplelist class itself, in the setter functions. Let me know if you don't understand anything.

saif-sk:
Hi, I would like to work on this. This will be my first attempt at open source contributing.

pragyasrivastava0805:
@apsdehal can I generate a PR for this?

pragyasrivastava0805:
@apsdehal @vedanuj could you please guide me and tell what exactly needs to be done?

PR #868, opened 2021-04-12
[feat] Device of tensor being added to SampleList is checked #567

prachigarg23:
The requirement was that when adding a tensor to an existing sample_list, the add_field method should check:

1. If a tensor field exists. 
2. The device of the tensor field.

The device of the tensor being added should be same as the device of the tensor field. If it isn't, migrate the data item to be added to the device of the tensor field. 

I have tested this in the case where:
I have a sample_list with device='cuda:0', I try to add data item with 'device=cpu' and the code successfully detects the device discrepancy and migrates the device of new data item to 'cuda:0'. 

Kindly let me know if my interpretation of the feature requirement is correct.

### DIFF
--- a/mmf/common/sample.py
+++ b/mmf/common/sample.py
@@ -312,6 +312,12 @@ class SampleList(OrderedDict):
                 )
             )
 
+        if tensor_field is not None:
+            device = self[tensor_field].device
+            if isinstance(data, torch.Tensor) and data.device != device:
+                if hasattr(data, "to"):
+                        data = data.to(device) # ensures the data being added is on the same device as existing tensors
+
         if isinstance(data, collections.abc.Mapping):
             self[field] = SampleList(data)
         else:


apsdehal:
Thank you for working on this and making MMF better. We are working on landing it to master.

prachigarg23:
Thanks @apsdehal for your response. I have 2 doubts:

1. I have made the changes (to check device and change device of new tensor) inside the add_field function itself. Should these changes have been made in the '_get_data_copy' method instead? I mean checking existing device and migrating data to the correct device. I'm confused since in the original feature description, you mention "SampleList in all cases, (update, setattr) should check...". Can you please confirm once if I am checking the device at the correct place?

2. _set_tensor_field (line 321 in master) will not be called for any new field other than the 1st field entered into the sample list. This is because line 320 requires tensor_field to be None in order to set a new field: 
`if isinstance(self[field], torch.Tensor) and tensor_field is None:`

I was wondering if this is intended?

apsdehal:
@prachigarg23 Your understanding is correct in both cases. I will leave some comments on PR.

Your linter (black) and unittests are failing at the moment, can you fix those?

mmf/common/sample.py
```diff
@@ -312,6 +312,13 @@ def add_field(self, field, data):
                 )
             )
 
+        if tensor_field is not None:
+            device = self[tensor_field].device
+            if isinstance(data, torch.Tensor) and data.device != device:
+                if hasattr(data, "to"):
+                    # ensures data being added is on same device as existing tensors
```
apsdehal, line 319: We should add a warning here and let user know that we are doing this.

tests/common/test_sample.py
```diff
@@ -100,6 +100,20 @@ def test_to_device(self):
         custom_batch = [{"a": 1}]
         self.assertEqual(to_device(custom_batch), custom_batch)
 
+        # Test if device of tensor being added inside add_field is compatible with existing sample_list tensors' device
+
+        # case 1: existing sample_list_ has device='cuda', we try to add a tensor with device='cpu'
+        sample_list_ = SampleList()
```
apsdehal, line 106: Can you create a separate test for this?

apsdehal (changes requested):

apsdehal:
@prachigarg23 Let me know if you are able to fix these tests otherwise I will take a look.

prachigarg23:
@apsdehal I shall fix these tests and make the changes, please give me time till this week end.

NoobMaster1999:
Is this issue still open? I want to contribute.

PR #872, opened 2021-04-13
[fix] Fixed drop_last setting and all_gather for xla device.

ultrons:
fixed drop_last setting for xla aligned with length computation.  fixed all gather (torch.stack only needed for GPU path.)


apsdehal:
Hi @ultrons, can you add `[fix]` tag ahead of your commit?

PR #913, opened 2021-04-29
[fix] CircleCI GPU tests; sharded DDP tests

apsdehal:
- Due to various reasons (num_workers=1) etc sharded ddp's process
destruction didn't work properly and gets stuck in between. This PR aims
to fix that.


vedanuj:
GPU tests are still failing it seems.

PR #921, opened 2021-05-03
[fix] multinode jobs after recent lightning update

apsdehal:
After Sasha's update of pytorch lightning on MMF master, it broke MMF codebase for multinode job. The root problem to PR https://github.com/PyTorchLightning/pytorch-lightning/pull/6802. The assumption that SLURM_PROCID points to worker rank is wrong as some frameworks launch their own processes later using multiprocessing spawn and have ntasks_per_node=1 set. This means that first node will have procid = 0, second node will have procid = 1 set and so on. Now, since this is used in prepare_data masking in LightningDataModule, this leads to it running on all workers on first node and thus causing inconsistencies. Now, this leads to prepare_data being called on all workers on first node instead of rank zero. Specifically, the barrier call in prepare_data, is called on first node workers but not on others leading to block later on.

This PR fixes this by ensuring on our side that we only call prepare_data on rank zero. Furthermore, this can cause further confusion, we remove sync barrier calls from download as well. Users are now supposed to handle is_master checks on their own.

Test Plan:
Tested in multinode settings.

### DIFF
--- a/mmf/datasets/base_dataset_builder.py
+++ b/mmf/datasets/base_dataset_builder.py
@@ -40,6 +40,7 @@ from typing import Optional
 import pytorch_lightning as pl
 from mmf.utils.build import build_dataloader_and_sampler
 from mmf.utils.distributed import is_master, synchronize
+from omegaconf import DictConfig
 from torch.utils.data import Dataset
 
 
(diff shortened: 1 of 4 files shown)


ytsheng:
Do you mind sharing a repro command that shows the breakage?

apsdehal:
@ytsheng You can try launching any sweep script on two nodes and it will get stuck at `Loading model` part.

PR #942, opened 2021-05-13
[fix] Only return cuda as current device if cuda is actually initialized

apsdehal:
This causes issues when GPU is available but person still want to run on
CPU.

Test Plan:

Tested locally


vedanuj (approved):
Looks good!

PR #954, opened 2021-06-01
[fix] fix TPU checkpoint saving, parameter broadcasting, and misc stuff

ronghanghu:
Fix the following on TPU training:
- instead of using `xm.save`, add a custom `save_xla_ckpt` to avoid
crashing on lr_scheduler state dict; **note that if we later switch back
to xm.save, we need to set `global_master=True` in xm.save**
- only delete old checkpoint from master process when using XLA
- only open file in "wb" mode from the master process when using XLA. (Otherwise, concurrent writing to the same file in `with PathManager.open(ckpt_filepath, "wb")` from multiple processes causes permission denied error).
- only save `config.yaml` from the global master process.
- skip batch size validation in XLA mode in evaluation loop (since data loader already automatically drops the last batch in XLA mode)
- On TPUs, there isn't an equivalent to DistributedDaraParallel yet, so adding a manual parameter broadcasting at the beginning of training (`broadcast_xla_master_model_param`). Otherwise we will have different parameters on each process since we are explicitly setting different seeds in different processes in MMF.
- set `drop_last=True` in DistributedSampler for XLA to correctly drop the last batch in XLA mode (the current codebase only set `drop_last=True` in DataLoader, which is not enough).

Test plan: tested on GCP.

### DIFF
--- a/mmf/utils/checkpoint.py
+++ b/mmf/utils/checkpoint.py
@@ -12,10 +12,11 @@ import torch
 from mmf.common.registry import registry
 from mmf.utils.checkpoint_updater import get_pretrained_state_mapping_checkpoint
 from mmf.utils.configuration import get_mmf_env, load_yaml
-from mmf.utils.distributed import is_master, is_xla, synchronize
+from mmf.utils.distributed import is_master, is_xla, open_if_master, synchronize
 from mmf.utils.download import download_pretrained_model
 from mmf.utils.file_io import PathManager
 from mmf.utils.general import get_current_device, updir
+from mmf.utils.xla import save_xla_ckpt
 from omegaconf import OmegaConf
 
 
@@ -489,7 +490,7 @@ class Checkpoint:
         }
 
     def save_func(self, *args):
-        return xm.save(*args) if is_xla() else torch.save(*args)
+        return save_xla_ckpt(*args) if is_xla() else torch.save(*args)
 
     def save(self, update, iteration=None, update_best=False):
         # Only save in main process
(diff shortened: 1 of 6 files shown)


mmf/utils/checkpoint.py
```diff
@@ -582,5 +585,39 @@ def restore(self):
 
     def finalize(self):
         if is_master() or is_xla():
-            with PathManager.open(self.pth_filepath, "wb") as f:
+            with open_if_master(self.pth_filepath, "wb") as f:
                 self.save_func(self.trainer.model.state_dict(), f)
+
+
+def save_xla_ckpt(ckpt, file_or_path):
```
apsdehal, line 592: This is awkwardly placed. Move to top or some other utils files such `xla.py` (create if not there).

mmf/utils/checkpoint.py
```diff
@@ -582,5 +585,39 @@ def restore(self):
...
+        )
+    else:
+        ckpt = xm._maybe_convert_to_cpu(ckpt, convert=should_write_data)
+
+    if should_write_data:
+        torch.save(ckpt, file_or_path)
+    xm.rendezvous("mmf.utils.checkpoint.save_xla_ckpt")
+
+
+def open_if_master(path, mode):
```
apsdehal, line 619: Same for this guy, move to utils/distributed.

apsdehal (approved):
This looks good to me!  Some comments around structuring and should be good to land.

ronghanghu:
I'll leave this PR open after confirming that SimCLR training is fully working on TPUs.

PR #959, opened 2021-06-05
[fix] set logging level according to training.logger_level

ronghanghu:
Correctly set logging level in the logger according to training.logger_level.

Also fix lint errors in other files to pass the tests.

Test plan: tested locally.

### DIFF
--- a/mmf/utils/logger.py
+++ b/mmf/utils/logger.py
@@ -98,10 +98,11 @@ def setup_logger(
     distributed_rank = get_rank()
     handlers = []
 
+    logging_level = registry.get("config").training.logger_level.upper()
     if distributed_rank == 0:
-        logger.setLevel(logging.INFO)
+        logger.setLevel(logging_level)
         ch = logging.StreamHandler(stream=sys.stdout)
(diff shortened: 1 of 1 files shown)


mmf/utils/logger.py
```diff
@@ -98,10 +98,11 @@ def setup_logger(
     distributed_rank = get_rank()
     handlers = []
 
+    logging_level = registry.get("config").training.logger_level.upper()
     if distributed_rank == 0:
-        logger.setLevel(logging.INFO)
+        logger.setLevel(logging_level)
```
apsdehal, line 103: Does it work directly, don't you have to do `getattr(logging, logging_level)`?

ronghanghu: I tested and verified that it works. In https://docs.python.org/3/library/logging.html#logging.Logger.setLevel, the input can be a str:
> The level parameter now accepts a string representation of the level such as ‘INFO’ as an alternative to the integer constants such as "INFO".

vedanuj (approved):
Looks good. Thanks for fixing!

PR #962, opened 2021-06-07
fix fp16 overflow issue

jayleicn:
"-1e9" induces "RuntimeError: value cannot be converted to type at::Half without overflow:" when fp16=true, replace it with "-10000.0"

Tested locally

Thanks for your contribution!


vedanuj:
Hi @jayleicn Please import the PR to Phabricator so we can run the internal tests.

Issue #976, opened 2021-06-18
mmf_interactive command


WesRobbins:
## 🐛 Bug

within intreractive.py, ```registry.get("config")``` returns None

## Command

mmf_interactive datasets=textcaps \
    model=m4c_captioner \
    config=projects/m4c_captioner/configs/m4c_captioner/textcaps/defaults.yaml \
    run_type=val \
    checkpoint.resume_zoo=m4c_captioner.textcaps.defaults \
    training.num_workers=0 \
    env.data_dir=/content/drive/MyDrive/data/mmf \
    env.cache_dir=/content/drive/MyDrive/cache/mmf

or 

from mmf_cli.interactive import interactive

ops_list = ['dataset=textcaps', 'model=mc4_cationer', 'training.num_workers=0', 'checkpoint.resume_zoo=m4c_captioner.textcaps.defaults'
            'env.data_dir=/content/drive/MyDrive/data/mmf', 'env.cache_dir=/content/drive/MyDrive/cache/mmf']
interactive(ops_list)

## Environment

PyTorch version: 1.8.1+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
CMake version: version 3.12.0

Python version: 3.7 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: Tesla P100-PCIE-16GB
Nvidia driver version: 460.32.03
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.0.4
[...]

PR #978, opened 2021-06-21
[fix] refactor regression

ytsheng:
A regression caused by a recent refactor. Fixes issue: https://github.com/facebookresearch/mmf/issues/976

### DIFF
--- a/mmf/utils/logger.py
+++ b/mmf/utils/logger.py
@@ -98,7 +98,12 @@ def setup_logger(
     distributed_rank = get_rank()
     handlers = []
 
-    logging_level = registry.get("config").training.logger_level.upper()
+    config = registry.get("config")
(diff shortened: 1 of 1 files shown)


mmf/utils/logger.py
```diff
@@ -98,7 +98,12 @@ def setup_logger(
     distributed_rank = get_rank()
     handlers = []
 
-    logging_level = registry.get("config").training.logger_level.upper()
+    config = registry.get("config")
+    if config:
+        logging_level = config.training.logger_level.upper()
```
apsdehal, line 103:
```suggestion
        logging_level = config.get("training", {}).get("logger_level", "info").upper()
```

apsdehal (changes requested):

PR #1002, opened 2021-07-05
[fix][PL][3/N] do not update loss in validation step

ytsheng:
* no need to update the loss in the validations tep
* simplify tests


ytsheng:
prefer https://github.com/facebookresearch/mmf/pull/1019

PR #1003, opened 2021-07-05
[feat][PL][4/N] PL best checkpoint (save/load)

ytsheng:
* Added monitor checkpoint for best metric
* Added test


ytsheng:
prefer https://github.com/facebookresearch/mmf/pull/1020

PR #1004, opened 2021-07-06
[fix] black

ytsheng:


apsdehal:
It looks like black and some other checks are still failing?

ytsheng:
The pillow issue is related to this: https://github.com/python-pillow/Pillow/pull/5572

PR #1009, opened 2021-07-09
[feat][PL][5/N] load model checkpoint directly + model zoo

ytsheng:
This PR tackles model only checkpoint loading, while previous PRs take advantage of the `resume_from_checkpoint` trainer param to resume the trainer state, this instantiates the model from the specified model only checkpoint. 

This test case covers loading a model checkpoint from the model zoo. The test case shows that loading model checkpoint from model zoo for both mmf and lightning results in the same model state_dict.

This PR also includes minor clean up of checkpoint functionality.


ytsheng:
prefer https://github.com/facebookresearch/mmf/pull/1021

PR #1027, opened 2021-07-21
[feat] check NaNs in losses during training, and exit when it happens

ronghanghu:
Add config option `training.exit_on_nan_losses` (Default to True).

When `training.exit_on_nan_losses` is True, check for NaNs in the loss dict and exit the training when NaN happens.

Test plan: tested locally; a new test case `test_exit_on_nan_losses` is added.

### DIFF
--- a/mmf/trainers/core/training_loop.py
+++ b/mmf/trainers/core/training_loop.py
@@ -164,9 +164,31 @@ class TrainerTrainingLoopMixin(ABC):
 
     def run_training_batch(self, batch: Dict[str, Tensor], loss_divisor: int) -> None:
         report = self._forward(batch)
+        if self.training_config.exit_on_nan_losses:
+            self._check_nan_losses(report)
         loss = extract_loss(report, loss_divisor)
         self._backward(loss)
         return report
+
+    def _check_nan_losses(self, report):
+        # skip this check in XLA mode as calling .item() in forward pass
+        # greatly slows down the training
+        if not is_xla():
+            # check whether NaN has occurred in the losses, and exit the training
+            # when NaN happens
+            loss_dict = report.losses
+            nan_loss_keys = []
+            for key, value in loss_dict.items():
+                if torch.any(torch.isnan(value)).item():
+                    nan_loss_keys.append(key)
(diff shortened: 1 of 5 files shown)


mmf/trainers/core/training_loop.py
```diff
@@ -164,10 +164,25 @@ def run_training_epoch(self) -> None:
 
     def run_training_batch(self, batch: Dict[str, Tensor], loss_divisor: int) -> None:
         report = self._forward(batch)
+        if self.training_config.exit_on_nan_losses:
+            self._check_nan_losses(report)
         loss = extract_loss(report, loss_divisor)
```
apsdehal, line 169: You can check afterwards on the loss as well, no? That would anyways be nan. I think ideal place for this check would be just before actual backward happens.

mmf/trainers/core/training_loop.py
```diff
@@ -164,10 +164,25 @@ def run_training_epoch(self) -> None:
 
     def run_training_batch(self, batch: Dict[str, Tensor], loss_divisor: int) -> None:
         report = self._forward(batch)
+        if self.training_config.exit_on_nan_losses:
```
apsdehal, line 167: Probably add a simple test for this following existing trainer tests if possible?

apsdehal (changes requested):

mmf/trainers/core/training_loop.py
ronghanghu: I was deliberately putting it over `report` rather than loss and before `extract_loss`, so that it would check for each loss before reduction, and provide some clue on which loss and which process lead to NaNs. But I can change it to just check the output of `extract_loss` if you feel that's better 😄

mmf/trainers/core/training_loop.py
ronghanghu: Sounds good!

ronghanghu:
A new test case `test_exit_on_nan_losses` is added.

vedanuj:
Looks good. Can you import to phabricator and run the internal tests. Thanks!

apsdehal (approved):
Please rebase before landing.

PR #1034, opened 2021-07-29
[fix] add `weight_decay` parameter to `get_bert_configured_parameters`

ronghanghu:
Allow specifiying custom weight decay in `get_bert_configured_parameters`. The behavior of existing codebase is kept the same for backward compatibility.

Test plan: tested in custom projects

### DIFF
--- a/mmf/utils/modeling.py
+++ b/mmf/utils/modeling.py
@@ -15,7 +15,7 @@ ACT2FN = {
 }
 
 
-def get_bert_configured_parameters(module, lr=None):
+def get_bert_configured_parameters(module, lr=None, weight_decay=0.01):
     # module param can either be a nn.Module or in some cases can also be
     # a list of named parameters for a nn.Module
     if isinstance(module, nn.Module):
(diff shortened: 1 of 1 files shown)


mmf/utils/modeling.py
```diff
@@ -15,7 +15,7 @@
 }
 
 
-def get_bert_configured_parameters(module, lr=None):
+def get_bert_configured_parameters(module, lr=None, weight_decay=0.01):
```
apsdehal, line 18: We should also update the call sites to use `weight_decay` from `config.optimizer.params.get("weight_decay", 0.01)` in case users are using this.

ronghanghu: > We should also update the call sites to use weight_decay from config.optimizer.params.get("weight_decay", 0.01) in case users are using this.

I'm a bit not sure here. I feel changing call sites to `config.optimizer.params.get("weight_decay", 0.01)` would be a backward-incompatible change and could break some existing MMF models or user models.

Issue #1036, opened 2021-07-31
VisualBert Doesn't return attentions


EXUPLOOOOSION:
## 🐛 Bug

the VisualBert model ignores the output_attentions config

## To Reproduce

In a python script:
1. get a configuration with output_attentions = True
2. Initialize and build any VisualBert model (any that uses VisualBertBase)
3. Run an inference (forward) and find out in the ouput that the attention tuple exists but is empty

specifying output_attentions to True in the forward function parameters doesnt help either.

## Expected behavior
Of course, attentions shouldn't be empty.

## Environment

PyTorch version: 1.9.0+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
CMake version: version 3.12.0
Libc version: glibc-2.26

Python version: 3.7 (64-bit runtime)
Python platform: Linux-5.4.104+-x86_64-with-Ubuntu-18.04-bionic
Is CUDA available: False
CUDA runtime version: 11.0.221
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.0.4
/usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.0.4
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.19.5
[pip3] pytorch-lightning==1.4.0
[pip3] torch==1.9.0
[pip3] torchmetrics==0.4.1
[pip3] torchsummary==1.5.1
[pip3] torchtext==0.5.0
[...]

PR #1052, opened 2021-08-22
[fix] VisualBERT returns attention tuple #1036

abhinav-bohra:
PROBLEM: The default value of output_attentions in forward( ) call of BertEncoderJit (in mmf/modules/hf_layers.py) is set as False. So even if the user/developer specifies output_attentions = True in config; its value is taken as default False and thus VisualBERT returns an empty tuple for attentions.

FIX: Set output_attentions as None in BertEncoderJit's forward( ) definition, and update output_attentions to self.output_attentions if it is not passed as an argument (i.e it is None). Therefore, now output_attentions will take the value of self.output_attentions (which was initialized using config during instantiation of BertEncoderJit class)

Same problem and same fix for output_hidden_states as well.

Tested locally.

Thanks for your contribution!

### DIFF
--- a/mmf/modules/hf_layers.py
+++ b/mmf/modules/hf_layers.py
@@ -280,11 +280,17 @@ class BertEncoderJit(BertEncoder):
         attention_mask: Optional[Tensor],
         encoder_hidden_states: Optional[Tensor] = None,
         encoder_attention_mask: Optional[Tensor] = None,
-        output_attentions: bool = False,
-        output_hidden_states: bool = False,
+        output_attentions: bool = None,
+        output_hidden_states: bool = None,
         return_dict: bool = False,
         head_mask: Optional[Tensor] = None,
     ) -> Tuple[Tensor]:
+        
+        if output_attentions is None:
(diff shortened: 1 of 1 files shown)


abhinav-bohra:
> 
> 
> Thanks for making this changes and contributing to MMF. Can you write a simple test to verify that this actually works. Test should go in `tests/models/test_visual_bert.py`.

Thanks for your feedback @apsdehal . I will add the tests.

I had one doubt - there can be 3 places where the user can set output_attention, namely;
1. In config
2. During instantiation of the model
3. During the forward pass

For example:
```
config = BertConfig.from_pretrained('bert-base-uncased', output_attentions = True)
model = VisualBERTBase(config, output_attentions = False)
y, z, attn = model.forward(self.x, output_attentions = True)
```
Should the priority be forward > model > config ?
i.e even if `output_attentions` is set as True in config but passed as False in forward, we should not return attention_list.

Or should it be an 'OR' of all the three cases; i.e. setting it anywhere as True will return the attention_list ?

PR #1058, opened 2021-08-27
Python version updated

syed-danish-in:


SyedDanish21nov:
Sorry for the late reply. But may I know why updated version of python 3.9
cannot be implemented?
May I know cons of updating, Please let me know if possible?
I'm waiting for your reply.

PR #1063, opened 2021-08-29
[feat] ALBEF Vision Transformer Encoder

PotatoSpudowski:
Summary:
PR adds ALBEF's Vision Transformer to MMF. 

- Adds Vision Transformer model
- Adds Vision Transformer encoder
- Adds tests for Vision Transformer encoder


mmf/models/vit.py
```diff
@@ -0,0 +1,486 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
apsdehal, line 1: I think we will probably delegate on HuggingFace to provide ViT implementation. Can you create a folder  `mmf/models/albef` and put this there. Instead of registering ViT encoder in `mmf/modules/encoders.py`, you can register the encoder `mmf/models/albef/vit.py` itself as `albef_vit_encoder` to be very specific.

apsdehal (changes requested):
Thanks for contributing to MMF. Let's first make this change to make it more specific to ALBEF.

mmf/models/vit.py
PotatoSpudowski: Hi @apsdehal thank you for your feedback. I will implement the requested changes.

apsdehal:
@PotatoSpudowski The linter tests are failing. Can you fix those?

mmf/models/__init__.py
```diff
@@ -16,6 +16,7 @@
 from .unimodal import UnimodalBase, UnimodalText, UnimodalModal
 from .visual_bert import VisualBERT
 from .vilbert import ViLBERT
+from .albef.vit import AlbefVitEncoder
```
vedanuj, line 19: We should only expose models here. Once the albef model is added you can add the import here.

mmf/models/albef/vit.py
```diff
@@ -0,0 +1,515 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+# Initial version was taken from https://github.com/rwightman/pytorch-image-models
```
vedanuj, line 3: We should try to avoid copy pasting so much code from `timm` library. It will be better to optionally import timm and just modify the necessary methods required in the vision transformer implementation for ALBEF model.

PotatoSpudowski: @vedanuj thank you for your feedback. I will make sure to fix these in the next PR when I add the ALBEF model.

mmf/models/albef/vit.py
```diff
@@ -0,0 +1,515 @@
...
+    def __init__(self, config: Config, *args, **kwargs):
+        super().__init__()
+        self.config = config.get("params", {})
+        pretrained = config.get("pretrained", False)
+        pretrained_path = config.get("pretrained_path", None)
+        self.vit = VisionTransformer(self.config)
+        if pretrained:
+            state_dict = torch.load(pretrained_path)
+            self.vit.load_state_dict(state_dict)
+            self.vit.eval()
```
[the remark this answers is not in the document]
PotatoSpudowski: Yup this seems like it will cause unintended behaviours. I will fix this while adding the remaining ALBEF model. Thank you for your feedback.

Issue #466, opened 2020-08-10
Multimodal Alignment loss for Vilbert/VisualBERT


vedanuj:
## 🚀 Feature
Multimodal Alignment or Sentence Image Prediction loss for ViLBERT/VisualBERT

## Motivation

[VilBERT](https://arxiv.org/abs/1908.02265) model uses two pretraining losses. Current MMF implementation uses `masked multimodal modeling` loss but not the `multimodal alignment` loss. Similarly [VisualBERT](https://arxiv.org/abs/1908.03557) model also uses a similar loss which they call  `sentence-image prediction` loss. Task is to add this multimodal alignment loss to Vilbert and VisualBERT models.

## Pitch

To be able to reproduce Vilbert/VisualBERT model results `multimodal alignment` loss should be added. Also in order to extend to retrieval downstream tasks, this `multimodal alignment` loss will be important.

## Additional context

The task will involve adding this loss to the models, modify any dataset side changes and test the implementation is working as expected.

Priyanshiguptaaa:
Hi @ytsheng, I would like to give this a try. Is there any other information or resources about this feature apart from this thread that I should know about?

Priyanshiguptaaa:
Hi @vedanuj @ytsheng, I would like to give this a try if the feature is still required to be worked on. If yes, is there any other information or resources apart from this thread that I need to know about? 

Thanks!

pragyasrivastava0805:
Hi @apsdehal I think I can do this. I will have to contribute this as a part of an initial screening for a research intern position. Hope you understand

Farhan-jafri:
Hi, I want to work on this issue, this is my first contribution, could you please guide me

PR #1065, opened 2021-08-31
Update vilbert.py

dk25021999:
Optional ITM loss in pre-training added for VilBERT.
Addresses #466

Thanks for your contribution!

### DIFF
--- a/mmf/models/vilbert.py
+++ b/mmf/models/vilbert.py
@@ -10,6 +10,7 @@ import torch
 import torch.nn.functional as F
 from mmf.common.registry import registry
 from mmf.models import BaseModel
+from mmf.models.transformers.heads.itm import ITM
 from mmf.modules.hf_layers import replace_with_jit
 from mmf.utils.configuration import get_mmf_cache_dir
 from mmf.utils.modeling import get_optimizer_parameters_for_bert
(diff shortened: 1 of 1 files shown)


mmf/models/vilbert.py
```diff
@@ -1226,6 +1230,14 @@ def forward(
                 prediction_scores_t.view(-1, self.vocab_size), masked_lm_labels.view(-1)
             )
             output["masked_lm_loss"] = masked_lm_loss.unsqueeze(0)
+        
+        if itm_loss is not False:
+            itm_head = ITM({"type": "itm", "hidden_size": self.vocab_size})
```
vedanuj, line 1235: This should be initialized in the init not in the forward pass.

vedanuj (changes requested):
Thanks for contributing to MMF. Requesting some changes before this is ready to merge. 

In addition, have you tried running this model? How have you tested this change?

mmf/models/vilbert.py
dk25021999: Hi! corrected the initialization thing. I checked the snippet I added separately in similar manner to "ITM head test" but was not able to test the ViLBERTForPretraining (As after loading yaml file its throwing 'dict' object has no attribute 'bert_model_name') even without making changes in original code.

PR #1072, opened 2021-09-01
[fix] Add zoo requirements to coco masked config

Ryan-Qiyu-Jiang:
Summary:
Add same zoo_requirements as coco dataset to masked_coco.
Prevents file not found error while loading datasets.


apsdehal (approved):
Thanks for the contribution! Looks good to me. Please import this to phabricator so we can land it.

PR #1073, opened 2021-09-01
[feat] add config option to skip update when NaN loss happens

ronghanghu:
Add config option `training.skip_batch_on_nan_losses`.

When setting to True, if the current batch results in NaNs in the losses, skip the optimizer (and grad scaler) update on this batch. This could be useful in avoiding occastional underflow or overflow when `training.fp16` is activated. Mutually exclusive with `training.exit_on_nan_losses`.

Also changing the previous nan checks from `isnan` to `isfinite` to also check for Inf.

Test plan: added new test case `test_skipping_batch_on_nan_losses`. Also tested locally in MMF training

### DIFF
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -153,11 +153,11 @@ training:
     # Set to true to look for NaNs in the losses and exit the training when NaN happens
     exit_on_nan_losses: true
 
-    # When setting to True, if the current batch results in NaNs in the losses, try
-    # forwarding on the previous batch in this iteration. This could be useful in avoiding
+    # When setting to True, if the current batch results in NaNs in the losses, skip the
+    # optimizer (and grad scaler) update on this batch. This could be useful in avoiding
     # occastional underflow or overflow when training.fp16 is activated.
     # Mutually exclusive with training.exit_on_nan_losses
-    try_previous_batch_on_nan_losses: false
+    skip_batch_on_nan_losses: false
 
 trainer:
     # Name of the trainer class used to define the training/evalution loop
--- a/mmf/configs/defaults.yaml
+++ b/mmf/configs/defaults.yaml
@@ -153,6 +153,12 @@ training:
     # Set to true to look for NaNs in the losses and exit the training when NaN happens
     exit_on_nan_losses: true
 
+    # When setting to True, if the current batch results in NaNs in the losses, try
+    # forwarding on the previous batch in this iteration. This could be useful in avoiding
+    # occastional underflow or overflow when training.fp16 is activated.
+    # Mutually exclusive with training.exit_on_nan_losses
+    try_previous_batch_on_nan_losses: false
+
 trainer:
     # Name of the trainer class used to define the training/evalution loop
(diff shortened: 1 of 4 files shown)


ronghanghu:
Closing as it didn't work on an actual training case (the training hangs upon NaN losses when using optimizer state sharding).

We can take an alternative route and handle this in the optimizer. For example, the following is a wrap of AdamW optimizer to handle NaN gradients.

```
import logging
import math
from typing import Callable, Iterable, Tuple

import torch
from transformers.optimization import AdamW


logger = logging.getLogger(__name__)


@registry.register_optimizer("adam_w_skip_nan_grad")
class AdamWSkipNaNGrad(AdamW):
    def step(self, closure: Callable = None):
        """
        Performs a single optimization step.

        Arguments:
            closure (:obj:`Callable`, `optional`): A closure that reevaluates the model and returns the loss.
        """
        loss = None
        if closure is not None:

            loss = closure()

        for group in self.param_groups:
            for p in group["params"]:
                if p.grad is None:
                    continue
                if not p.grad.isfinite().all().item():
                    logger.warning("Skipping param update due to NaN in gradients.")
                    return

        for group in self.param_groups:
            for p in group["params"]:
                if p.grad is None:
                    continue
                grad = p.grad.data
                if grad.is_sparse:
                    raise RuntimeError("Adam does not support sparse gradients, please consider SparseAdam instead")

                state = self.state[p]

                # State initialization
                if len(state) == 0:
                    state["step"] = 0
                    # Exponential moving average of gradient values
                    state["exp_avg"] = torch.zeros_like(p.data)
                    # Exponential moving average of squared gradient values
                    state["exp_avg_sq"] = torch.zeros_like(p.data)

                exp_avg, exp_avg_sq = state["exp_avg"], state["exp_avg_sq"]
                beta1, beta2 = group["betas"]

                state["step"] += 1

                # Decay the first and second moment running average coefficient
                # In-place operations to update the averages at the same time
                exp_avg.mul_(beta1).add_(grad, alpha=(1.0 - beta1))
                exp_avg_sq.mul_(beta2).addcmul_(grad, grad, value=1.0 - beta2)
                denom = exp_avg_sq.sqrt().add_(group["eps"])

                step_size = group["lr"]
                if group["correct_bias"]:  # No bias correction for Bert
                    bias_correction1 = 1.0 - beta1 ** state["step"]
                    bias_correction2 = 1.0 - beta2 ** state["step"]
                    step_size = step_size * math.sqrt(bias_correction2) / bias_correction1

                p.data.addcdiv_(exp_avg, denom, value=-step_size)

                # Just adding the square of the weights to the loss function is *not*
                # the correct way of using L2 regularization/weight decay with Adam,
                # since that will interact with the m and v parameters in strange ways.
                #
                # Instead we want to decay the weights in a manner that doesn't interact
                # with the m/v parameters. This is equivalent to adding the square
                # of the weights to the loss with plain (non-momentum) SGD.
                # Add weight decay at the end (fixed version)
                if group["weight_decay"] > 0.0:
                    p.data.add_(p.data, alpha=(-group["lr"] * group["weight_decay"]))

        return loss
```

PR #1075, opened 2021-09-07
[feat] Add metric support for dict return type

Ryan-Qiyu-Jiang:
Summary:
Add metric support for classes that return a dict of metrics.
All metric dictionaries are flattened in the final metrics dict.
Add example f1_precision_recall metric class as an example,
and for unit testing. This new metric class returns a dict,
containing the f1, precision, and recall tensors as values
with string keys.


tests/modules/test_metrics.py
```diff
@@ -145,6 +145,34 @@ def _test_retrieval_recall_at_k_metric(self, metric, value):
...
+                [-0.9332, 0.8149],
+                [-0.8391, 0.6797],
+                [-0.7235, 0.7220],
+                [-0.9043, 0.3078],
+            ],
+            dtype=torch.float,
+        )
+
+        metric_result = metric.calculate(sample, predicted)
+        print("metric_result", metric_result)
```
apsdehal, line 166: Remove debug warnings.

apsdehal (changes requested):
Thanks for the change. Remove print statements and we should be good to go!

apsdehal:
Also, rebase on master.

tests/modules/test_metrics.py
Ryan-Qiyu-Jiang: Done.

PR #1060, opened 2021-08-27
[feat] Weights and Biases Logging Feature

mudit-dhawan:
Add a WandbLogger class to support wandb logging. 
Update config options to enable wandb logger and integrate with the trainer.
To enable wandb logging use the following options `training.wandb=True  
env.wandb_projectname=<ProjectName> env.wandb_runname=<RunName>`

**Examples:**
Screenshot of the wandb dashboard for two runs of m4c project on textvqa dataset.  
image
1. command line code for _run1_: `mmf_run config=projects/m4c/configs/textvqa/defaults.yaml datasets=textvqa model=m4c run_type=train_val training.batch_size=2 training.log_interval=2 training.wandb=True training.max_updates=10 env.wandb_projectname=Try1 env.wandb_runname=run1 `
2. command line code for _run2_: `mmf_run config=projects/m4c/configs/textvqa/defaults.yaml datasets=textvqa model=m4c run_type=train_val training.batch_size=2 training.log_interval=1 training.wandb=True training.max_updates=10 env.wandb_projectname=Try1 env.wandb_runname=run2 `

 Linked Issue: Wandb Logger in mmf #588

PR #1088, opened 2021-09-11
[fix] Fix bug in wandb logging config logic

g-luo:
Although the docs state that the user should set `training.wandb.wandb_projectname` etc., the underlying code uses `config.trainer.wandb.wandb_projectname`. This small fix enables a user to successfully run the wandb logging feature. 

See #1060 for the original comment about the bug.


apsdehal:
@g-luo Thanks a lot for fixing this. We really appreciate it. The linter tests for black seem to be failing. Can you fix those?

PR #1129, opened 2021-10-20
[refactor] Extend WandbLogger to log config variables, entity and kwargs

ayulockin:
1. I was going through the `WandbLogger` class that was recently added to this library. Since MMF is a config-first framework, it would be great to log the config file to keep track of what's going in the model training/evaluation process. 

   The effect of this small change can be seen through the before and after screenshots. One can compare one experiment from another by comparing the config. 

   Before:
   image

   After: 
   image

2. I have also added another argument to pass `entity` to `wandb.init`. Explicit mention of this will help users who are using a W&B Teams account. 

3. Extra arguments to the `wandb.init()` can be passed via the config file. 

4. Ability to log learning rate over time. 

PS: I am learning about this library and will be using it for personal work. Since I am also affiliated with Weights and Biases, I will be creating some PRs to add functionalities to the logger that might be useful for everyone. :)

### DIFF
--- a/mmf/utils/logger.py
+++ b/mmf/utils/logger.py
@@ -2,7 +2,6 @@
 
 import collections
 import functools
-import itertools
 import json
 import logging
 import os
@@ -11,12 +10,14 @@ import time
 from functools import wraps
 from typing import Any, Callable, Dict, Optional, Union
 
+import omegaconf
 import torch
 from mmf.common.registry import registry
 from mmf.utils.configuration import get_mmf_env
 from mmf.utils.distributed import get_rank, is_main, is_xla
 from mmf.utils.file_io import PathManager
 from mmf.utils.timer import Timer
+from omegaconf import OmegaConf
 from termcolor import colored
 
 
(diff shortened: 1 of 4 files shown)


ayulockin:
Hey @ebsmothers is there any update on this? Would love to know if I need to make any changes. :)

ebsmothers:
Hi @ayulockin sorry for the delay and thanks for your patience. After a bit more consideration I decided we do not need to use open_dict for this case. To save you the trouble, I went ahead and updated the PR myself (with one other minor change). Thanks :)

mmf/utils/logger.py
```diff
@@ -426,17 +425,12 @@ def __init__(
...
-
-        wandb_params = config.training.wandb
-        with omegaconf.open_dict(wandb_params):
-            wandb_params.pop("enabled")
-            wandb_params.pop("entity")
-            wandb_params.pop("project")
-
-        init_kwargs = OmegaConf.to_container(wandb_params, resolve=True)
-        self._wandb_init.update(**init_kwargs)
+        self._wandb_init = dict(entity=entity, project=project)
```
ayulockin, line 428: Thanks for your prompt reply @ebsmothers and for making the change yourself. Is the minor change is in this line where `config=config` is removed from the dict?

Earlier I was passing the raw config dictionary to this `dict` so that it can be logged to W&B and one can reproduce an experiment from the logged configuration. 

Do you think it won't be of much use to the users? Happy to know your thoughts.

ebsmothers: Good catch. Actually the change I referred to was more minor, just a removal of .keys(). Config should still be included, so I added it back. Thanks for pointing out the omission

ayulockin: Yeah, I did notice the removal of `.keys()`. Thanks for adding back the config. So excited about how it progressed.

ayulockin:
Hey @ebsmothers, the change to `deepcopy(config.training.wandb)` threw an error: "DictConfig has no attribute pop" when `wandb_kwargs.pop("enabled")` was called so changed the type to `dict`.

ayulockin:
Closing this PR in favor of #1137.

PR #1178, opened 2021-12-15
[feat] Add a new VideoTransforms class

russellhowes:
Adds a new VideoTransforms class that will apply any transforms from pytorchvideo or torchvision.
All existing processors in video_processors.py can be replaced by one of these transforms.
Replaces references to existing processors in the Charades dataset config.
Once these references are replaced, the original processors can be cleaned up in a follow-up diff.
Adds unit test, fixes a unit test warning in test_processors.py about a collections.abc import
Changes a couple type annotations (mostly for my understanding; these functions may go away soon)


russellhowes (commented):
rebase

